PR #42, opened 2015-01-12
scheduling doc updates

jimmyjames:
reorganize a bit, add some more info around date string formats, and add info about discouraging using runIn to create schedules for less than sixty seconds.


smartapp-developers-guide/scheduling.rst
```diff
@@ -62,26 +63,43 @@ Executes the specified method in the specified number of seconds from now:
 
     You should not write your SmartApp expecting second-level granularity. You may find that it is accurate to the second-level in your case, but it cannot be guaranteed.
```
bflorian: Perhaps just say that the execution time may not be exact? The first time I read this section I interpreted *second-level granularity* as in, not *first-level granularity*, which of course is the wrong meaning.

PR #45, opened 2015-01-14
move groovy sandbox info to intro and add more information.

jimmyjames:
The section on groovy was buried in the IDE section. I moved it up to a first-level topic in the introduction, and included information about the Sandbox there, since it applies to both SmartApps and device-type handlers.

I also expanded the sandbox restrictions information a bit to enumerate some of the specific methods/properties that cannot be used, as well as list some other restrictions (e.g., builder limitations).


mrnohr:
:+1:

PR #46, opened 2015-01-16
Include some discussion of common groovy features

jimmyjames:


CosmicPuppy:
Great choice to include those particular Groovy examples. The web has a lot of resources, but it is hard to find ones that focus on relevant quirks and practical examples.

PR #50, opened 2015-01-23
update info on HubAction since it's been causing confusion.

jimmyjames:
@lbredeso you should review this before it gets merged (or ask others in the know to review)


lbredeso:
I'll take a look.

cloud-and-lan-connected-device-types-developers-guide/building-lan-connected-device-types/building-the-device-type.rst
```diff
@@ -1,71 +1,190 @@
...
+.. code-block:: groovy
+
+    def parse(description) {
+        ...
+        def msg = parseLanMessage(description)
+
+        def headersAsString = msg.header // => headers as a string
+        def headerMap = msg.headers      // => headers as a Map
+        def body = msg.body              // => request body as a string
+        ...
```
lbredeso: Other things that might be useful to detail in the return value of parseLanMessage(description):

status => HTTP status code of the response
json => Any JSON included in the response body
xml => Any XML included in the response body
data => Either JSON or XML included in the response body, which ever is specified by the Content-Type header in the response

cloud-and-lan-connected-device-types-developers-guide/building-lan-connected-device-types/building-the-device-type.rst
```diff
@@ -1,71 +1,197 @@
...
+.. code-block:: groovy
+
+    // gets the address of the hub
+    private getCallBackAddress() {
+        return device.hub.getDataValue("localIP") + ":" + device.hub.getDataValue("localSrvPortTCP")
+    }
+
+    // gets the address of the device
+    private getHostAddress() {
+        def parts = device.deviceNetworkId.split(":")
```
lbredeso: I'm conflicted about this example, because it assumes the device DNI (Device Network ID) is of the form "ip:port".  We really want to move away from that towards using the device MAC address as the DNI, and storing ip and port in device data.  WeMo (Connect) and the WeMo DeviceTypes use that pattern, for example.

On the other hand, if people have already developed their own DeviceTypes that assume ip:port DNI's, then this really is a valid example.

I guess outside of the specifics of this example, the important thing to note is that the device ip and port may not necessarily be stored in the DNI, and the developer just needs to know where to grab them from I guess.  :-)

PR #54, opened 2015-01-28
update state documentation

jimmyjames:


jimmyjames:
I think that's a good clarification we can add, maybe as a note in the state documentation. Preferences and State serve different purposes; this could be clarified in the documentation. 

I'll be out of pocket the rest of this week, but that's something I can easily do offline and get in early next week.

PR #56, opened 2015-02-04
add sunrise and sunset documentation

jimmyjames:
Added a section to the SmartApp Developer's Guide for Sunset and Sunrise. Also linked to some examples in the new Code repo in the SmartThingsCommunity project (we will want to more broadly publicize this at some point in the near future).

@mrnohr you should verify that the content is correct. @mager this is an example of linking to the examples from our guides.


mrnohr:
:+1:  Looks good, thanks.

PR #57, opened 2015-02-11
SmartApp dev guide events, device interaction updates.

jimmyjames:
Updates to our SmartApp documentation around events, subscriptions, and device interactions.

We had some information scattered around events, including a section on querying events, and one on generating events in SmartApps. I removed those and included information about subscription and events in the event-driven-smart-apps page. I kept that page to preserve bookmarks.

I also split out a separate section for interacting with devices in SmartApps, since this is common and felt like it deserved its own page.

@mrnohr you should review this or delegate as you see fit.


mrnohr:
:accept:

PR #61, opened 2015-02-18
unschedule is expensive. can't fix it quickly, so document it for now.

jimmyjames:
@mrnohr does it make sense to document the unschedule performance limitations for now? Take a look, thanks


mrnohr:
I think the documentation you added here is good

PR #63, opened 2015-02-23
Use corrected example (that works) for response()

garyd9:


Greblys:
Neither of these two examples work in device handlers. The new one is not functional at all, and the old one at least sends a bunch of commands at the same time, because delay is not functional, so you can't respond with too many commands.

PR #67, opened 2015-03-10
Update device handler guide

jimmyjames:
This is a rather large refactoring/expansion of the device handler developer's guide. 

It includes:
 - reorganization of content, including moving the Z-Wave and ZibBee device-type handler's guide out of the weird nested structure that exists today.
 - A new quick start for device handlers.
 - Dedicated pages for tiles, preferences (not complete yet), parse, and definition.

There's still more work to be done, obviously, but my hope is that this addresses the rather urgent need for this documentation to not be so incomplete.


mager:
I tested this device type quick start and it works!

PR #84, opened 2015-04-09
clarify overview to more accurately describe capabilities.

jimmyjames:


device-type-developers-guide/overview.rst
```diff
@@ -98,11 +98,13 @@ This code illustrates how a SmartApp might interact with a device that supports
         log.debug "switch is ${switch.currentSwitch}"
     }
 
-The above example illustrates how a SmartApp requests a device that supports the "Switch" capability. It can then work with the device knowing that it will support all the commands and attributes that the "Switch" capability supports.
+The above example illustrates how a SmartApp requests a device that supports the "Switch" capability. When installing the SmartApp, the user will be able to select any device that supports the "Switch" capability - be it an in-wall light switch, a connected bulb, a music player, or any other device that supports the "Switch" capability.
 
 There is a `reference
 document <https://graph.api.smartthings.com/ide/doc/capabilities>`__ that outlines all the supported capabilities.
 
+Device handlers typically support more than one capability. A device handler for a Hue bulb would support the "Switch" capability as well as the "Color Control" capability. This allows SmartApps to be written in a very flexible manner.
```
CosmicPuppy: Glad to see enhancements in Capabilities documentation.
I'm not sure what "typically" means in this context, however. Except for the "tagging" Capabilities (actuator and sensor), I think it could be argued that many, many SmartDevice Types actually only support one Capability. I would not consider a Motion Sensor or a Contact Sensor or a Temperature Sensor or a Switch, etc., etc., to be "atypical".

The Capability model certainly does enhance the flexibility of SmartApps, but this is not particularly due to "more than one capability" in Device handlers. So perhaps the statement of "flexibility" is misleading in this paragraph and belongs elsewhere.

device-type-developers-guide/overview.rst
```diff
@@ -75,7 +75,7 @@ Capabilities
...
 
-Consider the example of the "Switch" capability. There are many unique switch devices - in-wall switches, Z-Wave switches, ZigBee switches, etc. All these unique devices have a device handler, and the device handlers support the "Switch" capability. This allows SmartApps to only require a device that supports the "Switch" capability, and thus work with a variety of manufacturer and model-specific switches. 
+Consider the example of the "Switch" capability. In simple terms, a switch is a device that can turn on and off. It may be be a switch in the traditional sense (an in-wall light switch, for example), or a connected bulb (a Hue or Cree bulb, for example), or even a music player. All of these unique devices have a device handler, and those device handler's support the "Switch" capability. This allows SmartApps to only require a device that supports the "Switch" capability, and thus work with a variety of devices, including different manufacturer and model-specific "switches". The SmartApp can then interact with the device knowing that it supports the "on" and "off" command (more on commands below), without caring about the specific device being used.
```
CosmicPuppy: Lots of good information in this paragraph. I suggest that it *may* be useful to say that a Capability is actually specifically definable by its list of Attributes and Commands (provided those Attributes and Commands are well defined). The current text essentially says this in different words, of course. As per your example, if a Device can tell you if it is on or off (i.e., has a *binary power state* Attribute), and the Device can accept commands for changing its binary state (i.e., has `on()` and `off()` Commands), then that Device has Capability Switch.

It can even be written mathematically: Capability Switch = Attribute Switch + Commands( `on()`, `off()`); and the equivalency is fully reflexive ( A = B -> B = A ).

jimmyjames: If the device does not explicitly declare that it has a capability, SmartApps will not be able to leverage those attributes or commands in a flexible manner. It's possible to write a device handler that exposes the on() and off() commands, but if it does not explicitly declare that it supports the "Switch" capability, a SmartApp wouldn't be able to be installed for a "switch" input. 

But the capability is only meaningful through its attributes and commands (with the exception of the tagging capabilities), as that determines the commands and attributes available in a SmartApp, so I see your point.

device-type-developers-guide/overview.rst
jimmyjames: Many device handlers actually do support more than one capability. If you look through the device handlers code, you'll find that probably most do support more than one capability (not including the "tagging" capabilities). 

That said, the flexibility statement probably should be rearranged, since in this context it is one of the things that increase the flexibility, but not core to it.

CosmicPuppy: Cool. The term "typically" throws me off a bit, because I'm not sure if it means "most", "recommended", or just "frequently". There are certainly many device handlers with a single capability, and, depending on the granularity of the definition of future Capabilities, this could become more common or less common in the future. In discussions, I've suggested that highly-granular Capabilities definitions (i.e., as few Attributes and/or Commands as possible) are likely a *Best Practice*, in which case, indeed, it will be very typical for devices to have multiple capabilities. That condition does and will make coding the requests for `preferences{ input() }` developer and user unfriendly to some extent, though, as Aaron pointed out in this post:  http://community.smartthings.com/t/request-for-and-functionality-in-preference-input-of-smartapps/12768

device-type-developers-guide/overview.rst
CosmicPuppy: Absolutely. I guess I was suggesting that various terminology could have a "syntactical" and "technical semantic" definition in the Documentation in addition to natural language descriptions and examples. Technical math-like definitions are insufficient of themselves to explain a concept, but sometimes their *precision* can help clear up some ambiguities by providing an alternate form of expression.

PR #86, opened 2015-04-16
Added link to zwave util ref to references pg

surfous:
I always have to go to the z-wave primer to find this link (we don't need no stinkin' bookmarks!) and it's a reference that is also written/maintained/hosted by ST, so it seemed sensible to link to it from the reference materials page.


surfous:
Hey... first time with a pull request on the ST docs for a small mod. Let me know if I'm missing any conventions or process. Thanks.

PR #92, opened 2015-05-06
New ref docs

jimmyjames:
Major refactor of the reference docs. The current reference docs, hosted on graph.api.smartthings, poses a few issues:

 - Viewing the reference documentation requires a log in. Annoying.
 - Because it's password-protected, our documentation isn't indexed by the Google machine.
 - Because it's on a different server, the ReadTheDocs search doesn't include the reference documentation. Very inconvenient.
 - It creates unnecessary load on the graph.smartthings.api server (which is also where our IDE is hosted).
 
To remedy those issues, we are migrating our "reference" documentation into our documentation hosted by ReadTheDocs.

Our initial goals are:
1. Get the docs off of graph.api.smartthings.
2. Add clarifications/examples where it's easy to do so.

This initial roll-out does not include:
1. Documenting everything that's not currently documented.
2. Including examples for every API or capability.

We want to be iterative; get a good start out there, and incrementally improve upon it.

The Capabilities Taxonomy has also been moved into ReadTheDocs. It does not contain all documentation for all capabilities. It has the same "quick reference" info, as well as more information on some capabilities initially. This will take some time to fill in details about each capability, but we think we've got a good jumping-off point.

In the coming weeks we'll be redirecting the reference documentation on graph.api.smartthings.com back to the new documentation, and then removing it entirely. Update your bookmarks!


CosmicPuppy:
The improved convenience is wonderful news, Jim!

*But a slight concern...*
>In the coming weeks we'll be redirecting the reference documentation on graph.api.smartthings.com back to the new documentation, and then removing it entirely. Update your bookmarks!

I'm concerned that there may be "many" references in the **Community Discussion Forum** (http://community.smartthings.com) that link to the *old* URL's. Is there some way to arrange permanent automatic redirection; and/or search and replace of the moved links?

jimmyjames:
just a note not to merge this. There is still some work on this PR.

Regarding old, broken links, we can see if there's a way to permanently link to the new ones... but I'm not sure. It's a valid concern, and one that we will face more of in the future as we make changes, reorganize, or even consider a new documentation site/tooling altogether.

mager:
This might be an option: http://docs.readthedocs.org/en/latest/automatic-redirects.html

jimmyjames:
That may handle redirects in RTD, but in this case we need to redirect from graph.api.smartthings.com

PR #95, opened 2015-05-14
explicitly document that you cannot import your own classes/jars

jimmyjames:


CosmicPuppy:
Is there a list somewhere of the available classes that **SmartThings** has available for us to *import*? For example, "`import groovy.json.JsonSlurper`" works (and is quite useful), but I've not found a list of similar available classes on the platform. *Thanks!*

jimmyjames:
Nope. There is a white list of allowed classes that is not documented at this time. It's in the backlog.

PR #104, opened 2015-06-02
Updated SmartApp Publication Guidelines with @kris-schaller and Dwight

mager:
We wanted to improve the docs to give users more insight into the approval process.

### DIFF
--- a/device-type-developers-guide/submitting-device-types-for-publication.rst
+++ b/device-type-developers-guide/submitting-device-types-for-publication.rst
@@ -1,9 +1,6 @@
 Submitting Device Types for Publication
 =======================================
 
-Overview
---------
-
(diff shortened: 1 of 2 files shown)


alttext:
I'd like to see a Reasons For Publishing as another section that is a more positive counterpoint to reasons for rejection. It could illustrate some things we look for and get excited about. (new, unique, etc)

jimmyjames:
+1 @alttext. Would be good to let people know what we *do* look for, in addition to what we don't.

I'd also like to see a stand-alone guide/section for contributing that outlines the process, what we look for and what we don't, etc, for all types of submissions (versus a chapter in the various existing guides). That's beyond the scope of this single PR of course... but just thinking longer-term.

CosmicPuppy:
:+1:

PR #108, opened 2015-06-08
Update state

jimmyjames:
The current documentation for using state does not adequately discuss how it works, or discuss atomicState at all.

This PR includes:
* Updating guide title from "State" (implementation specific) to "Storing Data" (hopefully more meaningful to those who don't already know what "state" is).
* Add information about how state works at a high level (read from storage on initialization, and written back after execution completion).
* Add scenario of two concurrently executing installed SmartApps trying to use the same state variable.
* Discuss atomicState; why use it, when to use it, and how to use it.

/cc @beckje01 can you take a peek at the updates here?

### DIFF
--- a/smartapp-developers-guide/state.rst
+++ b/smartapp-developers-guide/state.rst
@@ -143,6 +143,8 @@ To avoid this type of scenario, you can use ``atomicState``. ``atomicState`` wri
 
   Use ``atomicState`` only if you are sure that using ``state`` will cause problems. 
 
+  It's also worth noting that you should **not** use both ``state`` and ``atomicState`` in the same SmartApp or Device Handler. Doing so will likely cause inconsistencies in in state values.
+
 Examples
 --------
 
--- a/smartapp-developers-guide/state.rst
+++ b/smartapp-developers-guide/state.rst
@@ -1,18 +1,65 @@
-State
-=====
+Storing Data
+============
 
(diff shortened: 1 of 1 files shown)


smartapp-developers-guide/state.rst
```diff
@@ -57,26 +108,40 @@ As usual, the best way to describe code is by showing code itself.
...
+
+.. code-block:: groovy
+
+  atomicState.counter = atomicState.counter + 1.
+
+.. important::
+  
+  Using ``atomicState`` instead of ``state`` incurs a higher performance cost, since external storage is touched on read and write operations, not just when the application is initialized or done executing.
+
+  Use ``atomicState`` only if you are sure that using ``state`` will cause problems.
```
beckje01: Also worth noting you should mix the two, if you modify some state variables and do atomic state the atomic changes will be overridden by the state changes during the execution since the state data structure isn't convergent.  @rappleg should confirm.

rappleg: As @beckje01 mentioned, you should NOT mix the two as that will cause issues with some state changes to possibly get overridden.

rappleg:
Other than a note about not mixing the two types of states this looks good to me.  Thanks for adding some more documentation around this, will be very helpful to folks.

:+1:

PR #109, opened 2015-06-09
Clarify when configure() is called

scottvlaminck:


jimmyjames:
Looks good - thanks for the clarification!

PR #111, opened 2015-06-10
add guide for sending notifications.

jimmyjames:
This is fairly straightforward; just adding some docs around sending notifications in SmartApps.

/cc @mrnohr @unixbeast


mrnohr:
:+1:

PR #112, opened 2015-06-10
Add info about sendNotification and sendNotificationEvent

jimmyjames:
Adding information about other notification APIs, `sendNotification()` and `sendNotificationEvent()`.

/cc @mrnohr @unixbeast


mrnohr:
:+1: looks good

PR #115, opened 2015-06-12
Cloud lan devices

unixbeast:


jimmyjames:
adding a comment for any public interest -

These are great changes that were identified and made as part of our review of the Cloud and LAN-Connected Devices guides. In no way to do we consider this set of documentation done; but they are relatively small changes with a high ROI. 

Nice work @unixbeast - those diagrams are awesome!

CosmicPuppy:
:+1: :+1: :+1:

PR #126, opened 2015-07-03
Add details for input element in device-handler preferences

CyrilPeponnet:


jimmyjames:
Thanks for helping to improve this @CyrilPeponnet. 

We will take a look at this next and test this out. We are hoping to better unify the preferences for devices and SmartApps to some extent, which should also allow us to better unify docs around prefs in general. It's still a somewhat moving target, so these docs haven't been fleshed out well yet.

Appreciate the contribution - some docs are better than no docs :) We'll follow up next week.

CyrilPeponnet:
This is far from perfect. You can add how to retrieve values with settings.xxx and also mention that it's a read only value. For sure listings devices per capabilities in a device handler doesn't make any sense but still allowed.

jimmyjames:
closing since preferences docs have been added to the device handler guide

PR #133, opened 2015-07-14
update docs with OAuth changes that impact getting the access token.

jimmyjames:
/cc @beckje01, @llinder, @unixbeast fyi. 

There's some formatting diffs included here you can ignore - content changes starts towards the bottom at "Get the API token" section heading.


beckje01:
:+1:

PR #139, opened 2015-07-22
updating zigbee example to reflect rename of Centralite Switch to Sma…

unixbeast:
updating zigbee example to reflect rename of Centralite Switch to SmartPower Outlet


jimmyjames:
image

PR #152, opened 2015-08-03
Simplify the docs organization. Add captions to the toc to categorize…

jimmyjames:
Simplify/clarify the docs organization.

This categorizes the docs into "getting started", "guides", and "reference". We can add additional categories as we go (e.g., cookbook, examples/tutorials, etc).

### DIFF
--- a/device-type-developers-guide/index.rst
+++ b/device-type-developers-guide/index.rst
@@ -1,15 +1,15 @@
 .. _device_type_dev_guide:
 
-Device Handler Developer's Guide
-=============================
+Device Handlers
+===============
 
-Device Handlers (sometimes also referred to as Device Type Handlers, or SmartDevices), are the virtual representation of a physical device. 
+Device Handlers (sometimes also referred to as Device Type Handlers, or SmartDevices), are the virtual representation of a physical device.
 
-If you are new to writing device handlers, start with the :doc:`quick-start`. 
+If you are new to writing device handlers, start with the :doc:`quick-start`.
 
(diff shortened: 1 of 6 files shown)


device-type-developers-guide/index.rst
```diff
@@ -1,15 +1,15 @@
 .. _device_type_dev_guide:
 
-Device Handler Developer's Guide
-=============================
+Device Handlers
```
alttext: Should this be the official name? Device Type Handler?

jimmyjames: When I updated this guide a few months ago, Bob said that the most accurate name is "Device Handler". But, this has caused confusion since people have generally adopted "Device Type Handler" (in the IDE they are called "SmartDevice").

alttext: (in the IDE they are called "SmartDevice")

Wha?

I have never heard that term.

CosmicPuppy: **Yes...**: The old version of this page (and related) documenting the API / Object Types used to refer to "SmartDevice Type" (I think the word "Handler" was never used, but I interpreted as meaning the "code" as opposed to the overall "type definition"). A Device was an instance constructed from a SmartDevice Type.

But I guess somewhere along the line the terminology changed to just "Device Type" ... or is it "Device Type Handler"? Not a bad thing if we get this straight; I thought it was a little awkward calling it a "*SmartDevice* Type", but I think that name came from the fact that we call an app object a "SmartApp", not an "App".

So ... what are the real Objects called now? I think consistent terminology is ***critical*** to avoid confusion! Thanks!!!

PR #156, opened 2015-08-06
Add reference docs for getting the API server URLs.

jimmyjames:
This adds some reference docs for getting the API server URLs.

/cc @kyleboon can you take a quick look?


kyleboon:
Looks good to me! :+1:

PR #160, opened 2015-08-17
clarify that atomicState is only available for SmartApps

jimmyjames:


jimmyjames:
This attempts to clarify that atomicState is only available in SmartApps currently, and adds reference docs for atomicState to Smartapp and state to Device handlers.

PR #161, opened 2015-08-18
Fixed HTTP Get error

pstuart:
HTTPGet was referencing HTTP DELETE not GET


jimmyjames:
copy/paste FAIL :)

Thanks for catching and fixing!

PR #166, opened 2015-09-15
only single quotes can be used to interpolate string vals in the tiles definition

jimmyjames:
A community member was (understandably!) confused that string interpolations in the tiles method use single quotes, and fails with double quotes. This is not as Groovy's string interpolation is defined (only works with double quotes), so causes confusion.

I added a clarifying tip near this code example that was causing confusion. What we really need is an entire lifecycle diagram/overview as discussed previously, but given the current docs backlog I don't want this clarification to go undocumented until such time as we find bandwidth/priority for the lifecycle diagram.

@mrnohr would be good for you to review/comment/suggest.


mrnohr:
Agreed that it is confusing. Good addition. :+1:

PR #188, opened 2015-10-27
Update division-of-labor.rst

ughodke:
fixing typos


jimmyjames:
good catch, thanks!

ughodke:
Thanks, I guess not many walk these (Device Handler) pages...   ;-)    I'll keep posting as I progress.  I have also drawn up a system diagram that may need to be reviewed by ST for posting. Thanks,Umesh

PR #189, opened 2015-10-30
Fix broken Z-Wave response() example

mckeed:
It doesn't appear to work to return only a HubAction from parse(), it only processes hub actions that are returned as part of a list. Returning [event, response(cmds)] is the current best practice for this kind of handler.


jimmyjames:
Thanks @mckeed, appreciated!

PR #191, opened 2015-11-04
updating web services smartapp example to use the URI parameter

unixbeast:
updating web services smartapp example to use the URI parameter

### DIFF
--- a/smartapp-web-services-developers-guide/tutorial-part2.rst
+++ b/smartapp-web-services-developers-guide/tutorial-part2.rst
@@ -279,14 +279,14 @@ Replace the ``/getswitch`` route with the following:
       puts json
 
       # get the endpoint from the JSON:
-      endpoint = json[0]['url']
+      uri = json[0]['uri']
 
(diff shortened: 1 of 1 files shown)


jimmyjames:
:+1: 

It's hard for me to be incremental in this change; I'd like to see docs that explain each endpoint in the OAuth process, and all of the returned values for each. BUT, until we can get priority/bandwidth to do that, keeping these updated to reflect the current changes is a must.

Can you update the example Sinatra app here too? https://github.com/SmartThingsCommunity/Code/blob/master/smartapps/tutorials/web-services-smartapps/server.rb

Thanks!

jimmyjames:
I wonder if it would be helpful if we put an admonition (`.. important:`) at the top of this page with a blurb about this recent change, to draw attention to it? Also we will probably need to clarify that any SmartApp self-published will only be available for installation via web flow to that specific region (e.g., won't be available to all regions unless published by ST)?

Either way this commit is good to go, we can talk about those things later.

PR #197, opened 2015-11-09
Correct typo in "input" section

mitchpond:
Correct cacapability to capability in "input" section, "type" table


jimmyjames:
:+1: nice find, thanks.

PR #204, opened 2015-11-18
New capabilities

Yaima:
- Sound sensor


Yaima:
@jimmyjames @tslagle13

jimmyjames:
Thanks Yaima! 

The only thing I'd ask is that we put the sound sensor in alphabetical order in the "At a Glance" table, rather than just at the bottom.

Other than that it loks good!

tslagle13:
:+1: after change Jim mentioned

jimmyjames:
:100: thanks!

PR #213, opened 2015-11-30
oauthPage preferences

jimmyjames:
add documentation for using oauthPage option for web services preferences.

### DIFF
--- a/smartapp-web-services-developers-guide/smartapp.rst
+++ b/smartapp-web-services-developers-guide/smartapp.rst
@@ -25,6 +25,83 @@ These will be used on the SmartThings Authorization page to inform the user who 
 
 ----
 
+Preferences
+-----------
+
+Part of the :ref:`Authorization Flow <webservices_authorization>` that installs the SmartApp requires the user to authorize specific devices that the third party can interact with.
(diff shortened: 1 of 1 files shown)


unixbeast:
do you think you should call this out specifically?
"Additional pages cannot contain any inputs - only display information (e.g., for additional information, instructions, etc.)"

jimmyjames:
Yes I think something is needed around that. More research is needed on my end to know if this is really the case, or if there's a way to make it work currently.

This will have to be separate from the ongoing questions/gaps of how the auth page handles preferences differently, including what features are planned to work, and what is not. Unfortunately given current priorities and capacity, I don't know think these questions will be answered any time soon. 

Will update this PR with updates later.

jimmyjames:
Updated PR to include:
* basic auth preferences using single preferences page
* specifying `oauthPage` in preferences for oauth-specific preferences
* using dynamic pages for additional, non-device inputs

@llinder I'd like you to review this prior to merging, to ensure that this documentation is aligned with the intent/design/functionality of the authorization page.

llinder:
The documentation looks correct based on my understandings of its intended use :+1:

PR #214, opened 2015-11-30
clarify that contact book is not an end-user available feature yet.

jimmyjames:
While contact book is not available for end users yet, developers should still use the contact book APIs (with the fall-back mechanism documented) to ensure their SmartApp is future-proofed.


unixbeast:
:+1:

PR #216, opened 2015-12-04
Correct authorization header in endpoints call example

jlinn:


jimmyjames:
nice find, thanks! :+1:

PR #221, opened 2015-12-11
Fixed minor wording mistake

Hanse00:
It appears like a minor error in wording snuck itself into the doc, as SmartThings probably ship with SmartApps, not with SmartThings.


unixbeast:
:+1: 
Thanks Hanse00!

Hanse00:
Always happy to help!

PR #225, opened 2015-12-21
Action was being overwritten by overloaded usage

StrykerSKS:
In the page2() dynamicPage, the setting "action" was being used twice and over written.


jimmyjames:
:+1:

PR #237, opened 2016-01-12
Edit IDE logging location reference

stevenhaddox:


jimmyjames:
I don't think we want to remove the information about where to find logging in the IDE for installed SmartApps, which is under "Live Logging". 

That said, maybe there should be something added parenthetically here about the logging available when using the simulator (which is the logging pane below the editor when a SmartApp is installed via the IDE simulator)?

stevenhaddox:
Oh, is this information for installed smart apps? I was following along in the tutorial and hadn't yet seen a step to install, or if I had then I hadn't seen the mentioned link in the appropriate place. Certainly don't remove that if it's accurate and I apologize if I misunderstood the intent of the logging link being mentioned.

stevenhaddox:
Submitted a PR with more accurate verbiage based upon feedback. Thanks for the clarification :)

PR #239, opened 2016-01-13
Clarify when "Live Logging" link will exist

stevenhaddox:
As a newb to SmartThings development following this tutorial I couldn't see the referenced "Live Logging" link. This is because you haven't actually installed a SmartApp yet at this point in the tutorial. Hoping this will clarify for other new community members as well.


stevenhaddox:
Do'h - I finally saw this link all the way in the top menu. I was looking in the IDE section. #fail

PR #240, opened 2016-01-13
DOCS-224: removing GET as a method option for getting access token

unixbeast:
removing GET as a method option for getting access token


iamcanadian2222:
@unixbeast @jimmyjames it might help to clarify the specifications required for the POST formatting:

eg. `x-www-form-urlencoded` and an example of the body `grant_type=authorization_code&code=YOUR_CODE&client_id=...`

jimmyjames:
:100:

PR #245, opened 2016-01-20
Document state character size limitation (DOCS-211)

jimmyjames:
The amount of data that can be stored in state or atomicState is currently limited to 100K characters. This PR documents that limitation.

I also added a Best Practices section that summarizes some of important information to keep in mind when using state.

@rappleg I think you should check this for accuracy before it's merged.


rappleg:
Looks good.  :+1:

PR #249, opened 2016-01-29
Correct the parameter input type for setThermostatX commands

### LINKED ISSUE #247: is enum the correct parameter type for setThermostatFanMode and setThermostatMode?
When looking at the capabilities here for a Thermostat: http://docs.smartthings.com/en/latest/capabilities-reference.html is uses enum as the input type for setThermostatFanMode and setThermostatMode. However, it seems that these should really be type String instead since you can only set one mode at a time for each of these.

I can make the changes to the documentation and do the pull request, but wanted to verify that my assumption is correct or not.

StrykerSKS:
This corrects the input type from "enum" to "string"

Closes #247


jimmyjames:
:+1:

PR #251, opened 2016-02-07
DOCS-249 - add new capabilities to docs

jimmyjames:
@Yaima since you added these capabilities, would like your review before merging. Thanks!

/cc @jodyalbritton

### DIFF
--- a/capabilities-reference.rst
+++ b/capabilities-reference.rst
@@ -51,6 +51,7 @@ Commands:
 :ref:`battery`                capability.battery                     - battery
 :ref:`beacon`                 capability.beacon                      - presence
 :ref:`button`                 capability.button                      - button
+:ref:`c_d_measurement`        capability.carbonDioxideMeasurement    - carbonDioxide
 :ref:`c_m_detector`           capability.carbonMonoxideDetector      - carbonMonoxide
 :ref:`color_control`          capability.colorControl                - hue                                 - setHue(number)
 
@@ -89,6 +90,7 @@ Commands:
                                                                                                            - resumeTrack(string)
                                                                                                            - restoreTrack(string)
 :ref:`notification`           capability.notification                                                      - deviceNotification(string)
+:ref:`ph_measurement`         capability.pHMeasurement               - pH
 :ref:`polling`                capability.polling                                                           - poll()
 :ref:`power_meter`            capability.powerMeter                  - power
 :ref:`presence_sensor`        capability.presenceSensor              - presence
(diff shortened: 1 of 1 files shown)


capabilities-reference.rst
```diff
@@ -51,6 +51,7 @@ Commands:
 :ref:`battery`                capability.battery                     - battery
 :ref:`beacon`                 capability.beacon                      - presence
 :ref:`button`                 capability.button                      - button
+:ref:`c_d_measurement`        capability.carbonDioxideMeasurement    - carbonDioxide
```
Yaima: I recommend using carbonDioxideMeasurement instead of carbonDioxide.

Yaima:
CO2 is returning a number  and CO is an enum. We need to keep measurement  for this CO2 cap. @jimmyjames @jodyalbritton

capabilities-reference.rst
```diff
@@ -109,6 +112,8 @@ Commands:
 :ref:`switch`                 capability.switch                      - switch                              - on()
                                                                                                            - off()
 :ref:`switch_level`           capability.switchLevel                 - level                               - setLevel(number, number)
+:ref:`sound_pressure_level`   capability.soundPressureLevel          - soundPressureLevel
+:ref:`tamper_alert`           capability.tamperAlert                 - tamper
```
Yaima: Instead of tamper, what do you think about tamper alert? @jimmyjames @jodyalbritton

capabilities-reference.rst
jimmyjames: Not sure I follow - isn't the attribute name `carbonDioxide`? That's what I see in StandardCapabilities.

capabilities-reference.rst
jimmyjames: I'm ok with `tamper` as the attribute name.

jimmyjames:
Maybe it would help to differentiate between any discrepancies between this docs PR and any changes we want to make for these capabilities? I just wanted to doc these capabilities as they exist now. I definitely don't want to doc it wrong so if they are wrong, we need to correct them now.

If they are still in active development or the names/attributes/commands will change soon, let's list those and not document them until they are stable and ready to be doc'd and supported.

jimmyjames:
@Yaima see my comments; we can chat tomorrow if that's easier. Thanks!

capabilities-reference.rst
jodyalbritton: I agree with carbonDioxideMeasurement

jimmyjames:
@Yaima per our conversation yesterday, let's decide if we should either document these capabilities now as they exist, or hold off on the docs if we want to finalize on those names (for the capabilities themselves, the attribute name, and command names).

We should either close this PR pending that decision, or merge/update as needed to be correct per the capability definitions.

Thanks!

Yaima:
Let's merge now because these capabilities are already deployed and bootstrapped. :+1:

PR #257, opened 2016-03-16
DOCS-271 - scheduling docs updates

jimmyjames:
Updating the scheduling docs to be better organized, as well as updated with any information around the ticker rollout.

There isn't anything too ticker-specific here, just an opportunity to clarify and expand on the scheduling docs.

@charliek I'd appreciate your review before I merge this.

/cc @tslagle13


jimmyjames:
and if you'd like to see/review as the docs will look, ping on slack/hipchat and I can get you details to view the docs as they will look live, instead of just text diff's.

smartapp-developers-guide/scheduling.rst
```diff
@@ -61,304 +69,358 @@ Executes the specified method in the specified number of seconds from now:
...
-Oftentimes, there is a need to schedule a job to run on a specific schedule. For example, maybe you want to turn the lights off at 11 PM every night. SmartThings provides the *schedule* method to allow you to create recurring schedules.
 
-The various *schedule* methods follow a similar form - they take an argument representing the desired schedule, and the method to be called on this schedule.
-Each SmartApp or device-type handler can only have one handler method scheduled at any time. This means that, unlike *runIn* or *runOnce*, a job created with *schedule* must either execute or be canceled with the *unschedule* method before you can schedule another job with the same method. The *schedule* method does not accept the overwrite option like *runOnce* and *runIn*.
-
-**schedule(dateTime, handlerMethod)**
+.. note::
 
-Creates a scheduled job that calls the handlerMethod every day at the time specified by the dateTime argument. The dateTime argument can be a String, Date, or number (to schedule based on Unix epoch time).
+    Unlike ``runIn()`` or ``runOnce()``, there is no option to overwrite existing scheduled executions.
```
charliek: With ticker if you schedule a cron job and one exists it will overwrite the current one and just change the interval, assuming you are scheduling the same method name.  There is no option to overwrite or not however since I didn't see how that would make sense for cron.

I thought I tested this in the simulator and just duplicated how it worked now, and didn't intend to break any existing contract.  We could change this if we wanted, but I actually like just re-scheduling more myself.

smartapp-developers-guide/scheduling.rst
```diff
@@ -61,304 +69,358 @@ Executes the specified method in the specified number of seconds from now:
...
+SmartThings allows you to pass a cron expression to the schedule method to accomplish this.
+A cron expression is based on the cron UNIX tool, and is a way to specify a recurring schedule.
 
-    def someEventHandler(evt) {
-        // call handlerMethod every day, at two minutes from the current time
-        schedule(now() + 120000, handlerMethod)
+.. code-block:: groovy
+    :emphasize-lines: 3
+
+    def initialize() {
```
charliek: I'd love it if people would use random seconds in schedules if they think they are going to be high volume and that is acceptable for their use case.  Should we document that at all you think?

smartapp-developers-guide/scheduling.rst
```diff
@@ -61,304 +69,358 @@ Executes the specified method in the specified number of seconds from now:
...
-Other Scheduling-related Methods
---------------------------------
+----
 
-**canSchedule()**
+Removing Scheduled Executions
+-----------------------------
 
-returns ``true`` if a job can be scheduled, ``false`` otherwise. Only four jobs may be scheduled for the future at any time.
+You can remove scheduled executions using the :ref:`smartapp_unschedule` method:
```
charliek: Don't know if we want to point out this will cancel both cron and run in schedules? A minor thing that I tripped on initially.  Ok to leave it out since sometimes less is better.

smartapp-developers-guide/scheduling.rst
```diff
@@ -61,304 +69,358 @@ Executes the specified method in the specified number of seconds from now:
...
-        canSchedule()
+    def handler() {
+        log.debug "in handler, current time is ${new Date()}"
     }
 
+You can call ``unschedule()`` with no arguments to remove all schedules.
+
+.. note::
+
+    Due to the way that the scheduling service is currently implemented, ``unschedule()`` is a fairly expensive operation, and may take many seconds to execute.
```
charliek: I swear it will get better once we get ticker 100% rolled out :)  But lets not put that into writing for the community just yet.

charliek: This should get faster with ticker, but will remain this way until we can get that 100% rolled out.  Probably best to leave this as is if that is what you've observed.  Kills me it is so slow.

smartapp-developers-guide/scheduling.rst
```diff
@@ -61,304 +69,358 @@ Executes the specified method in the specified number of seconds from now:
...
+We plan to address this in the near future. Until we do, be aware of the potential performance impacts of calling ``unschedule()``.
 
-As discussed above, ``unschedule`` is currently a potentially expensive operation.
+Note that when the SmartApp is uninstalled, all scheduled executions are removed - there is no need to call ``unschedule()`` in the ``uninstalled()`` method.
 
-We plan to address this in the near future. Until we do, be aware of the potential performance impacts of calling ``unschedule``.
+Number of Scheduled Executions Limit
+````````````````````````````````````
+
+The :ref:`smartapp_can_schedule` method returns false if four or more scheduled executions are created.
```
charliek: We would make a community post on this before enforcing a new limit, unless we had to do a critical change to avoid a system issue in which case it may come after.  But we are trying hard to be better around getting posts out.

smartapp-developers-guide/scheduling.rst
```diff
@@ -15,23 +15,26 @@ Broadly speaking, there are a few different ways we might want to schedule somet
...
 - If a door opens and is not shut after five minutes, send a notification.
 
-All these scenarios follow a common pattern: when a certain event happens, take some action after a given amount of time. This can be accomplished this by using the *runIn* method.
-
-**runIn(seconds, method, options)**
+All these scenarios follow a common pattern: when a certain event happens, take some action after a given amount of time.
+This can be accomplished this by using the :ref:`smartapp_run_in` method.
 
-Executes the specified method in the specified number of seconds from now:
+``runIn()`` takes the number of seconds from now to execute, the method to execute, and a map of options (not required).
```
charliek: Once ticker is 100% rolled out and we don't think we have to rollback for any reason we should discuss and document the new data option.  This basically allow you to pass data through a schedule, so for example you could do a run in and pass context for the followup execution. This will help avoid any race conditions that might happen in state values.  In general avoiding state where possible is a good best practice and this will give you one additional option for that.  The option will be present for both runin and cron jobs.

charliek:
Other than the cron overwrite comment this looks good to me and that is pretty minor.  My other comments are more just context and things to look out for in the future.  Thanks for getting this out there.

+1

smartapp-developers-guide/scheduling.rst
```diff
@@ -61,304 +69,358 @@ Executes the specified method in the specified number of seconds from now:
...
 
-- "Once a Day" uses ``schedule`` to turn switches on and off every day at a specified time.
-- "Turn It On For 5 Minutes" uses ``runIn`` to to turn a switch off after five minutes.
-- "Left It Open" uses ``runIn`` to see if a door has been left open for a specified number of minutes.
-- "Medicine Reminder" uses ``schedule`` to check if a medicine door has been opened at a certain time.
+.. [1] You may notice that some of the scheduling APIs accept a string to represent the the date/time to be executed.
+    This is a result of when you define a preference input of the "time" type, it uses a String representation of the value entered.
+    When using this value later to set up a schedule, the APIs need to be able to handle this type of argument.
+    When simply using the input from preferences, you don't need to know the details of the specific date format being used.
+    But, if you wish to use the APIs with string inputs directly, you will need to understand their expected format.
```
charliek: Do people actually use this string method?  Seemed really odd to me that we even have it.  But it should still work.

smartapp-developers-guide/scheduling.rst
jimmyjames: I can clarify this further to just state that cron overwrites existing scheduled methods.

smartapp-developers-guide/scheduling.rst
jimmyjames: I think that makes sense. I'll add some info around that.

smartapp-developers-guide/scheduling.rst
jimmyjames: Yes, better to be explicit, will update.

smartapp-developers-guide/scheduling.rst
jimmyjames: People do, though not common. Might be because that's what is documented, but we should make it clear the intent of the method is to take the value from a `time` input

smartapp-developers-guide/scheduling.rst
```diff
@@ -61,304 +69,373 @@ Executes the specified method in the specified number of seconds from now:
...
 .. note::
 
-    Full documentation of the cron expression format can be found in the `Quartz Cron Trigger Tutorial <http://quartz-scheduler.org/documentation/quartz-1.x/tutorials/crontrigger>`__.
-
-.. important::
+    High volume cron schedules are encouraged to specify a random seconds field.
+    This helps to avoid a large number of scheduled executions being queued up at the same time.
 
-    Cron jobs will only be allowed to run at a minimum of 1 minute intervals.
+    It may not be acceptable for every use case, but if it is, use a random second.
```
jimmyjames: Added a note to recommend using a random seconds field in cron if applicable to the use case.

smartapp-developers-guide/scheduling.rst
jimmyjames: added a note to suggest this below.

smartapp-developers-guide/scheduling.rst
jimmyjames: I clarified this in the latest commit

jimmyjames:
Thanks for the review @charliek! I updated the PR to address your line items.

PR #264, opened 2016-03-21
document job history may not be available as we migrate to new scheduler

jimmyjames:
@jodyalbritton @charliek


charliek:
The history not being available is actually not related to the migration at all.  The reason we turned it off is to shed load on cassandra as we tried to troubleshoot load issues with SHM.  We actually turned history back on this morning, since the load is looking a lot better and it is very useful for troubleshooting scheduler issues.

jimmyjames:
Thanks @charliek looks like my info was incorrect, I'll close this PR. Good to hear it's back on because it is super useful to have, and confusing when it's not there.

PR #267, opened 2016-03-22
using main tile with a list is useless now

jimmyjames:
Passing a list of tiles to `main()` is no longer useful, since the user cannot choose their main tile. Removing it from the docs to avoid confusion.


tslagle13:
:+1:

PR #268, opened 2016-03-24
global variables seem like they should work, but don't

jimmyjames:
Seen this before, and there was some chatter over in the community slack around how to create global variables in a SA/DT. Some apps are defining global constants, which compile fine, but are always `null`.

This PR attempts to clarify how to create a "global" constant.

@mrnohr maybe you can review if not we can bother a cloud dev


mrnohr:
:+1: Seems right to me

jimmyjames:
Just found an inconsistency around global vars in the docs; will update the PR shortly.

PR #270, opened 2016-03-29
DOCS-274 enumerate default response/error conditions for web service handlers

jimmyjames:
This PR adds some information around the default responses sent back for the various request methods, as well as errors that may be thrown by the platform.

@rappleg you or someone of your delegating should review to ensure this is consistent with your expectations, and also so you are aware should you want/plan on changing anything.

/cc @mrnohr


mrnohr:
:+1:

PR #275, opened 2016-04-01
formatting web services guide

jimmyjames:
Also reordered structure so the tutorials come before the other documentation.

@unixbeast


unixbeast:
:+1:

PR #282, opened 2016-04-01
fix build warnings

jimmyjames:
This gets us down to 1 pesky warning about the favicon not existing; it is being picked up, so it's not clear to me yet how to avoid the build warning but still get the favicon.

@unixbeast 

I think we can address these warnings and spend a little time later to see if we can get rid of the one remaining warning. Would be great to get a totally clean build.


unixbeast:
:+1:

PR #283, opened 2016-04-01
add style guide link to README

jimmyjames:
also added a fix for a separator here, since I noticed it.

@unixbeast


unixbeast:
:+1:

PR #284, opened 2016-04-01
SHARD-103: Rewrite of Service Manager documentation

lbredeso:
There's a lot of changes in here. Here's a summary at a high level:

1. Promote use of new ssdpTerm location event subscriptions
2. Show limited example of HubAction callbacks
3. Include better information on UPnP, especially linking to the actual specification, not Wikipedia articles (I like Wikipedia, don't get me wrong, but it has dramatically less information in this case)
4. Include Generic UPnP Service Manager and Device examples, for illustrative purposes (they also actually work)

CC-ing folks who have asked me about this work, hope I haven't forgotten anyone.

/cc @unixbeast @jimmyjames @larsfinander @juano2310 @Yaima @CCarden @tslagle13

### DIFF
--- a/cloud-and-lan-connected-device-types-developers-guide/building-lan-connected-device-types/building-the-service-manager.rst
+++ b/cloud-and-lan-connected-device-types-developers-guide/building-lan-connected-device-types/building-the-service-manager.rst
@@ -1,225 +1,267 @@
+============================
 Building the Service Manager
 ============================
 
-The Service Manager's responsibilities are to:
+The Service Manager's responsibilities are:
 
-- Discover devices
-- Handles device Add/Change/Delete actions
-- Maintains the connection
+- **Discovery** - Discover devices on the LAN via SSDP (mDNS/DNS-SD or Bonjour is not currently supported)
+- **Verification** - Verify each discovered device through successful fetching of the UPnP device description
+- **Inclusion** - Add the device as a child of the service manager
+- **Health** - Track IP address and port changes and allow these to make it down to the child device(s) as necessary
 
-Let's take a look at an example of what is outlined above.
+Let's take a look at some of the key parts of a Service Manager implementation. The example code referenced throughout
+this document is derived from the below SmartApp and DeviceType:
+
+:download:`Generic UPnP Service Manager <src/generic-upnp-service-manager.groovy>`
+:download:`Generic UPnP Device <src/generic-upnp-device.groovy>`
+
+The above referenced Generic UPnP Device is incomplete. For the complete guide, please see `Building the Device Type <building-the-device-type.html>`_.
+
+----
 
 Discovery
 ---------
 
-SSDP
-~~~~
+Simple Service Discovery Protocol (SSDP) is the main protocol used to find devices on your network. It serves as the backbone of Universal Plug and Play (UPnP), which
+allows you to easily connect new network devices to a system. See `UPnP Device Architecture 1.1 <http://upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.1.pdf>`__
+for full specification details.
 
-`SSDP <http://en.wikipedia.org/wiki/Simple_Service_Discovery_Protocol>`__
-is the main protocol used to find devices on your network. It serves as
-the backbone of `Universal Plug and
-Play <http://en.wikipedia.org/wiki/Universal_Plug_and_Play>`__, which
-allows you easily connect new network devices to a system. To discovery
-new devices, you'd use something like this:
+To discover new devices, you first need to subscribe to location events with the correct **search target** for the device. The
+search target in the below example, **urn:schemas-upnp-org:device:ZonePlayer:1**, is for discovery of a Sonos, but search targets will vary by manufacturer and device.
+For UPnP, this information should be published on documentation for the device, but you may
+alternatively have to contact the manufacturer directly to obtain it. Here is the event subscription:
 
 .. code-block:: groovy
 
-    sendHubCommand(new physicalgraph.device.HubAction("lan discovery urn:schemas-upnp-org:device:ZonePlayer:1", physicalgraph.device.Protocol.LAN))
+    subscribe(location, "ssdpTerm.urn:schemas-upnp-org:device:ZonePlayer:1", ssdpHandler)
 
-.. note:: sendHubCommand and HubAction are supplied by the SmartThings framework
+This means that any time an SSDP **search response** with a search target of urn:schemas-upnp-org:device:ZonePlayer:1
+(e.g. Sonos) is received from a hub in this location, it will fire the ssdpHandler method.
+
+Next, you need to send an appropriate discovery command for the desired search target:
+
+.. code-block:: groovy
+
+    void ssdpDiscover() {
+        sendHubCommand(new physicalgraph.device.HubAction("lan discovery urn:schemas-upnp-org:device:ZonePlayer:1", physicalgraph.device.Protocol.LAN))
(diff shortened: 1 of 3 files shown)


pstuart:
"Avoid making calls from your child devices into the parent if possible, as this causes unnecessary platform load; supply your child devices with enough information to make calls into the parent unnecessary, and use the Service Manager to manage any child device updates that need to happen based on network changes."

This seems to go against all existing examples and practices for building connection managers and simple devicetypes.  I would love to know more what extra load this causes by having the logic in the parent instead of the devicetype?

It also creates many fragments of code siting in each devicetype, especially for 3rd party multiple device support, where a variety of simple devicetypes that just hold the state of the device and a UI layer but the parent smartapp does all the communication.

It could also cause many devices to hit the "hub" (not smartthings) async and flood it with requests.

For example, if all the logic to communicate the the hue bridge was in the hue bulb, each bulb would be sending commands to/from the devicetype, rather than the hue smartapp?  Maybe routing those to the hue bridge makes sense, but since there is no method to go to/from a devicetype to devicetype, the data has to be routed to the parent smartapp.

Also, as a code example, doesn't it make more sense to use a switch statement for this code block:

`+private def parseEventMessage(String description) {
 +	def event = [:]
 +	def parts = description.split(',')
 +	parts.each { part ->
 +		part = part.trim()
 +		if (part.startsWith('devicetype:')) {
 +			part -= "devicetype:"
 +			event.devicetype = part.trim()
 +		} else if (part.startsWith('mac:')) {
 +			part -= "mac:"
 +			event.mac = part.trim()
 +		} else if (part.startsWith('networkAddress:')) {
 +			part -= "networkAddress:"
 +			event.ip = part.trim()
 +		} else if (part.startsWith('deviceAddress:')) {
 +			part -= "deviceAddress:"
 +			event.port = part.trim()
 +		} else if (part.startsWith('ssdpPath:')) {
 +			part -= "ssdpPath:"
 +			event.ssdpPath = part.trim()
 +		} else if (part.startsWith('ssdpUSN:')) {
 +			part -= "ssdpUSN:"
 +			event.ssdpUSN = part.trim()
 +		} else if (part.startsWith('ssdpTerm:')) {
 +			part -= "ssdpTerm:"
 +			event.ssdpTerm = part.trim()
 +		} else if (part.startsWith('headers')) {
 +			part -= "headers:"
 +			event.headers = part.trim()
 +		} else if (part.startsWith('body')) {
 +			part -= "body:"
 +			event.body = part.trim()
 +		}
 +	}
 +	event
 +}`

Super nested if / then /else can be very confusing for people to follow and it also has no default handler if part doesn't match.

Just my two cents.

juano2310:
@lbredeso you can use parseLanMessage(description) which is already in the platform and avoid defining private def parseEventMessage(String description)

lbredeso:
"This seems to go against all existing examples and practices for building connection managers and simple devicetypes. I would love to know more what extra load this causes by having the logic in the parent instead of the devicetype?"

I realize not all Service Managers are currently constructed in the pattern of this example code. Hue (Connect), for example, looks to currently have all Hue Bulb commands routed through it. Every time there is an "on" command sent to a Hue Bulb, then, the entire Groovy sandbox has to be spun up for both Hue Bulb, and for Hue (Connect) (the "parent" call). It works, of course, but is not as efficient as just directly issuing an HTTP call to the bridge itself from the Hub Bulb (via HubAction, I mean). I think it doesn't necessarily have to be the way it currently is.  There is data currently in Hue (Connect) that could live in the Hue Bulbs, for example. Each piece of data they need could be stored in device data (including the bridge IP), and they could then function independently of Hue (Connect). Maybe I'm missing details, and it isn't 

For that ```parseEventMessage``` method, yes, that's gross; I'll change that to something better.

lbredeso:
I replaced parseEventMessage.

juano2310:
:+1:
Completely unrelated. Should we move convertHexToInt and convertHexToIP to the platform?

pstuart:
I think it would be great to add these as core platform functions, Considering I can still find improper converting of Hex to IP from when I fixed it almost 2 years ago to handle 2 digit first octet ip addresses (ex, 10.x.x.x)

As for the model, the reality is the model works for ip connected devices, where their ip is unique via discovery or manual input.  Would love to see a generic manual input of ip example to be included as well.  Since there currently no way to get MAC from an ip / request (outside of SSDP discovery) you have to ask the end user to enter the MAC for DNI creation along with ip and port.

Since not all IP devices support SSDP, it would be nice to have docs to show how to handle this scenario and a model for when the device has only one IP but spawns many virtual child devices, where the parent smartapp needs to manage or route commands.

yracine:
Hello,

Based on my experience with building DTHs, I would like to chip in.

For MyEcobee, MyAutomatic, and MyNeurio DTHs and their corresponding Service Manager, I've followed the following design principles:

- The Service Manager is responsible for discovering the child devices, and maintaining the auth tokens (i.e. refreshing them when necessary).

- The child devices are as independent from the their parent as possible.  All commands are only implemented by the child devices. There is no back and forth between the parent and its child devices to execute a command.  

- No state or atomicState variables are used to store data between the parent and child devices (except for the auth tokens). The child devices rely on their own data.* structures and process them on their own (without the parent's help).

In my book, a DTH is a component that should be autonomous as much as possible  from its parent (Service Manager) to avoid any latency/concurrency issues...The DTH implements all the required event handlers and can respond to any event that concerns it.  

This autonomous design could be also expanded later to allow further decoupling between the parent and its child devices; allowing later some DTH local processing even though  its Service Manager must reside in the cloud (for cloud-to-cloud integration).

My 2 cents...

pstuart:
Yeah, for wan / cloud devices this is possible, and ideal.  Too bad no lan or wan / cloud DTH's run local though.

lbredeso:
@juano2310 @pstuart convertHexToInt and convertHexToIP should definitely exist somewhere in the platform. I'll log a ticket to do that.

@pstuart manual IP input is tricky. Asking for an IP is not ideal, but asking for both IP and MAC would be super annoying. I don't have any great ideas on how to improve this off the top of my head, and also don't have the bandwidth to write documentation on that particular scenario. This document was pretty SSDP-focused when I started, so I left that focus as-is.

@yracine increased latency is definitely a reason to avoid child -> parent calls if possible.

yracine:
Hi @lbredeso, I'm glad that somebody at ST is validating my design... I was kind of the only ST community developer  with such a design based on my observations of other Service Managers out there.

As soon as you create complex DTH with more than 10 commands and thousands of lines of code, the usual Service Manager design out there has a lot of limitations as you need to store a lot of data in state or even atomicStates variables which is not optimal.  The constant back and forth between the child devices and the parent is burdensome and does not provide any added value. It also complicates the code and affects its readability and maintenance.

On top of it, with the latest DB issues (which started in December) that created a lot of latency between the parent and its child devices, my DTHs were able to run independently despite all the ST platform problems.  

In brief, the autonomous design helped me to work around the platform issues in the past months.

For more details on MyEcobee, My Automatic, and My Neurio DTHs, please refer to the ST
community wiki:

http://thingsthataresmart.wiki/index.php?title=My_Ecobee_Device
http://thingsthataresmart.wiki/index.php?title=My_Automatic_Device
http://thingsthataresmart.wiki/index.php?title=My_Neurio_Device

Regards.

P.S. I sent you a linkedIn invitation to connect further.

lbredeso:
Per @unixbeast's suggestion, I added some minor style corrections from here: http://docs.smartthings.com/en/latest/contributing/style-guide.html

PR #294, opened 2016-04-07
DOCS-277 - do not allow arbitrary command execution through an endpoint

jimmyjames:
I think at some point in the near future some of this needs to be moved to, and expanded on, in a dedicated best practices section or page. For now we want to get this out there soon so it's not leading people down a wrong path, and provides some clarity on not allowing arbitrary command execution through endpoint mappings.

@danlieberman @unixbeast

### DIFF
--- a/smartapp-web-services-developers-guide/tutorial-part1.rst
+++ b/smartapp-web-services-developers-guide/tutorial-part1.rst
@@ -139,20 +139,18 @@ If any of the configured switches does not support the specified command, we'll 
         // use the built-in request object to get the command parameter
         def command = params.command
 
-        if (command) {
-
-            // check that the switch supports the specified command
-            // If not, return an error using httpError, providing a HTTP status code.
-            switches.each {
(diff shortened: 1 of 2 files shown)


smartapp-web-services-developers-guide/tutorial-part1.rst
```diff
@@ -139,20 +139,18 @@ If any of the configured switches does not support the specified command, we'll
...
-                if (!it.hasCommand(command)) {
-                    httpError(501, "$command is not a valid command for all switches specified")
-                }
-            }
-
-            // all switches have the comand
-            // execute the command on all switches
-            // (note we can do this on the array - the command will be invoked on every element
-            switches."$command"()
+        // all switches have the comand
```
lbredeso: Should be "command".

jimmyjames: thanks, good catch. Added a new commit to fix

unixbeast:
:+1:

PR #295, opened 2016-04-12
DOCS-281 - remove logging console from docs

jimmyjames:
@jodyalbritton this removes references to the logging console from the docs as discussed.


jimmyjames:
@unixbeast can you review? I think I got all references to using the console logging window, but I know you looked at logging docs some time ago so maybe you're aware of something hiding out there. Thanks.

unixbeast:
This looks good to me. I think I'll still poke around a bit to see why this is an issue to begin with.

PR #296, opened 2016-04-13
adding clarification around dotted expressions and groovy strings.

unixbeast:


jimmyjames:
image

sjurgemeyer:
:+1:

jimmyjames:
Maybe we should just recommend to always use `${}` - it's always safe and saves the confusion.

dsainteclaire:
👍

PR #299, opened 2016-04-14
Revert "Documented the new single-phase ZigBee device discovery process"

jimmyjames:
This reverts commit 4d75bf3b3a3526f37c5db70aca0cbd03b4d68b13.

@tpmanley @unixbeast I accidentally triggered a docs build by fat-fingering in the admin section of our docs site. This revert just makes it so we can rebuild and not have this out until the next FW release. I'll take care of that revert when needed.

Also, it's beyond time to move to a release branch and avoid this kind of caretaking :angry:


unixbeast:
:+1:

PR #300, opened 2016-04-14
DOCS-269 update images for new app UX

jimmyjames:
Images in the device tiles page will come as part of the overall work to fix that page, which is up next (woot!)


unixbeast:
:+1:

PR #303, opened 2016-04-18
setLevel clarification

jimmyjames:
While internal discussion around this command is ongoing, I at least want to provide _some_ level of clarification in the docs until it's all sorted out.


unixbeast:
👍

larsfinander:
Awesome

iamcanadian2222:
Now if only the devices would report that they only support one parameter for `setLevel` when looking at their `arguments` from `device.supportedCommands`! :stuck_out_tongue_winking_eye:

PR #310, opened 2016-05-10
Remove "default" from the description

StrykerSKS:
The word "default" in the description description can cause confusion leading the developer believe that it works similar to defaultValue which (in hind sight) is not the case.


jimmyjames:
👍

PR #311, opened 2016-05-13
DOCS-291 - updating the return value for /api/smartapps/endpoints

unixbeast:


jimmyjames:
👍 to merge pending the DM deploy

tslagle13:
Thanks guys!

jimmyjames:
@unixbeast I think we should explain what the list of returned json objects is and why it's a list.

unixbeast:
@llinder could you please answer the question that Jim posted in the above comment?

llinder:
@jimmyjames correct /api/smartapps/endpoints will return a list of all installed SmartApps for the client ID associated with the given access token.

jimmyjames:
👍 thanks guys

iamcanadian2222:
It's officially documented as a list/array response, so that means we can use this for multi-location authorization now, right? :stuck_out_tongue_winking_eye: @625alex

PR #312, opened 2016-05-13
making corrections to Timed Session and Power capabilities

unixbeast:


capabilities-reference.rst
```diff
@@ -1506,7 +1506,8 @@ power           String  ``"unknown"``
 
 **Commands:**
 
-None.
+*powerSource(String)*
```
jimmyjames: `powerSource` is the attribute, not a command :)

jimmyjames:
comment on `powerSource` inline. Once that's fixed up (just change the name of the attribute from `power` to `powerSource`), 👍 and go ahead and merge and build.

jimmyjames:
💯

PR #316, opened 2016-05-24
Cli

unixbeast:
First draft stage of CLI docs


jimmyjames:
👍 for first-stage. I have some formatting/organization feedback but will follow up with another PR

PR #335, opened 2016-06-24
DOCS-310 breaking out music player capability

unixbeast:
breaking out music player capability into 3 capabilities:
* audio notification
* music player
* tracking music player


iamcanadian2222:
Will `musicPlayer` remain for the foreseeable future or should we look at porting our DTHs over to the other capabilities?

PR #338, opened 2016-06-29
[DOCS-311] - clarify endpoint URLs vary by location

jimmyjames:
@unixbeast please review

/cc @jodyalbritton


unixbeast:
Looks good to me. should probably have @jodyalbritton verify it meets the community concerns.

tslagle13:
Could we add something that tells people how to get there "shard" without going through the process? Don't know if that makes sense, but that's normally the biggest ask.

jimmyjames:
Not sure I understand @tslagle13... what do you mean by "their shard"? The only way developers should ever get the endpoint of a SmartApp is to use the /endpoints API as documented, or the API Endpoint url in the simulator (if installing in the simulator to test).

jimmyjames:
@tslagle13 any clarifications per my comment you can add? Perhaps you are referring to the generic (non-web services SmartApp) desire to know what shard they are on, for just general information purposes? We should get a ticket created to discuss this generically (but not use the word "shard"). But for web services stuff if you have more info/clarity that would be great, thanks!

jodyalbritton:
Looks good to me

tslagle13:
Maybe my comment lives somewhere else than here. It could be argued that it is more of a customer thing. But many times customers are asked what "shard" they are on and don't know how it get that information.

jimmyjames:
I think there are two areas that need to be addressed around "shards":
1. Customer-facing KB articles targeting end users to understand what shard they are on.
2. Developer-facing documentation discussing the "sharding" architecture.

I'll create a DOCS ticket for 2), @tslagle13 can you work with support to get something on their radar for 1)?

Regardless, I will merge this PR and address future "sharding" docs in a separate PR.

PR #339, opened 2016-07-05
[DOCS-304] added documentation around using collections in state

jimmyjames:
Using collections in atomicState has caused confusion for developers when trying to modify the collection (adding, changing, or removing elements). The behavior is different between `state` and `atomicState`, creating more confusion.

This PR attempts to document the differences and a recommended way of working with collections in `atomicState`. 

I do not know if this behavior is intentional, a bug, or an unintended consequence of the implementation of `atomicState`. This PR simply documents how things behave right now.

@rappleg or @Bijnagte I'd like you to review/confirm this, or delegate to someone who can review.

/cc @dsainteclaire

### DIFF
--- a/smartapp-developers-guide/state.rst
+++ b/smartapp-developers-guide/state.rst
@@ -15,7 +15,6 @@ Recall that SmartApps (and Device Handlers) are not always running, but rather e
 Here's a quick example showing how to work with state:
 
 .. code-block:: groovy
-    :linenos:
 
     state.myData = "some data"
     log.debug "state.myData = ${state.myData}"
@@ -47,7 +46,6 @@ The contents of ``state`` are stored as a JSON string. This means that anything 
   This is particularly worth noting when working with dates. If you need to store time information, consider using an epoch time stamp, conveniently available via the ``now()`` method:
 
   .. code-block:: groovy
-    :linenos:
 
     def installed() {
       state.installedAt = now()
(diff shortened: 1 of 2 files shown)


smartapp-developers-guide/state.rst
```diff
@@ -189,6 +186,49 @@ When the character limit has been exceeded, a ``physicalgraph.exception.StateCha
...
+
+    def initialize() {
+        state.myMap = ["key1": "val1"]
+        log.debug "state: $state"
+        state.myMap.key1 = "UPDATED"
+        log.debug "state: $state" // state is now [key1: UPDATED]
+    }
+
+Updating collections stored in Atomic State is a little trickier - following the same pattern as above **will not work**.
+Instead, you will need to assign the collection to a local variable, make changes as needed, then assign it back to ``atomicState``.
```
Bijnagte: Actually that should work it is just a no no. It will call back to the database on each read and on each assignment. That means that if they are going to use the atomicState multiple times they should read it once and assign it to a local variable to avoid multiple reads.

Bijnagte:
As soon as Doug's changes to InstalledSmartAppState are active in production then we will also have the idea of convergence on first level keys. It will probably be easiest to explain in person but that should be documented as well.

smartapp-developers-guide/state.rst
jimmyjames: That's what I would have thought, but it doesn't actually work - the values are never updated in the collection. That's the primary source of confusion encountered by developers with `atomicState`.

jimmyjames:
DOCS-315 will address the first-level convergence on keys in state. After talking to Dylan I have a good understanding of that upcoming change.

I'd like to proceed with this PR for now, since this targets how to work with collections stored in state and atomicState, something that has tripped up others.

unixbeast:
👍

PR #345, opened 2016-07-11
[DOCS-314] add grails and groovy version to docs

jimmyjames:
Documenting the grails/groovy version used.

@Bijnagte please review

### DIFF
--- a/getting-started/groovy-for-smartthings.rst
+++ b/getting-started/groovy-for-smartthings.rst
@@ -13,6 +13,20 @@ Before we discuss the specifics of what is and what is not available to your Sma
 
 ----
 
+Groovy Version
+--------------
+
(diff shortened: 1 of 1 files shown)


Bijnagte:
I would say that I don't know that it matters to end users that we run grails since they are sandboxed away from that anyway. At some point in the not to distant future we may not be running executions from within grails at all.

jimmyjames:
I would tend to agree, but there are certain cases where grails classes/impacts leak to the developer APIs. It's not great, but is what it is, which is why I opted to include it for transparency. I could add a note to clarify that grails isn't something that should directly impact developers to help clarify...

Bijnagte:
Sure, I don't see any big issues then, I just am not sure we expose that we run grails anywhere officially. Looks correct to me though

Bijnagte:
👍 once it reflects reality (grails upgrade still in progess)

PR #357, opened 2016-07-30
[DOCS-319] remove sept 2015 release faq from top nav

jimmyjames:
Also moves the docs change log up to the top of the nav tree.

/cc @tslagle13


tslagle13:
:+1:

PR #359, opened 2016-08-01
event.dateValue does not throw exception if not parseable, it just re…

pfrybar:
…turns null. Fix the docs to reflect this. Also fix a few typos.


jimmyjames:
nice 👍

PR #360, opened 2016-08-02
[DOCS-254] - add links in the Device Handler guide to other available APIs

jimmyjames:
Device Handlers can call scheduling, state, and http APIs (for the most part, with exceptions noted in the changes in this PR). Our current docs organization makes discovering these APIs within Device Handler development less than optimal. Rather than duplicate the content within the DH guide, or not do anything until the entire organization/information hierarchy is redone, I just created a new page with links to topics in the SmartApp guide.

I also changed a couple titles of docs to make them more clear/discoverable.


unixbeast:
👍

PR #364, opened 2016-08-03
[DOCS-324] Start of troubleshooting docs for OAuth SmartApps

jimmyjames:
We probably should have had some troubleshooting docs for OAuth SmartApps for a while, and I think there's more that can be added here. But there are some topics I think we can address to start, that will be helpful for developers who get stuck in common areas.

@llinder would like your review of this before merging.

/cc @unixbeast @tslagle13


llinder:
👀

jimmyjames:
@llinder any thoughts on this PR?

llinder:
👍

PR #382, opened 2016-08-25
CREX-5898 - Document how to send data to a scheduled handler.

pfrybar:
CC @charliek @bflorian @jimmyjames @rappleg 

Documenting how to send data to a scheduled handler method. This is a good feature, which could help prevent people from abusing the state. Note: this only works with the cloud scheduler, and not the local one.

screen shot 2016-08-25 at 11 18 57 am

### DIFF
--- a/smartapp-developers-guide/scheduling.rst
+++ b/smartapp-developers-guide/scheduling.rst
@@ -279,6 +279,37 @@ Using them is similar to other scheduling methods:
 
 ----
 
+Passing Data to the Handler Method
+--------------------------------------
+
+Sometimes it is useful to pass data to the handler method. This is possible by passing in a map as the last
+argument to the various schedule methods with ``data`` as the key and another map as the value.
+
+.. code-block:: groovy
+    :emphasize-lines: 2
+
+    def someEventHandler(evt) {
+        runIn(60, handler, [data: [flag: true]])
+    }
+
+    def handler(data) {
+        if (data.flag) {
+            theswitch.off()
+        }
+    }
+
+By passing data directly to the handler method, you can avoid having to store data in the SmartApp or Device Handler state. Data can be passed
+in this fashion to any of the schedule methods: ``runIn()``, ``runOnce()``, ``schedule()``, ``runEveryXMinutes()``, and ``runEveryXHours()``.
+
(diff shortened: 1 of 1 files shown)


pfrybar:
Should wait to merge until PR 3105 from DM goes through, which changes the exceeded length behavior.

charliek:
We should wait until that PR makes it to production before merging this, as opposed to just waiting for it to be merged.

smartapp-developers-guide/scheduling.rst
```diff
@@ -279,6 +279,37 @@ Using them is similar to other scheduling methods:
 
 ----
 
+Passing Data to the Handler Method
+--------------------------------------
+
+Sometimes it is useful to pass data to the handler method. This is possible by passing in a map as the last
+argument to the various schedule methods with ``data`` as the key and another map as the value.
+
```
charliek: Should we say anything about how everything in the map must be able to be serialized to json, so adding something like a device wrapper or something may cause an error (or unexpected behavior).  This should be similar to how the state variables work as well.

charliek:
Thanks for getting this documented @pfrybar. It is something we couldn't document when we were on multiple schedulers, and it slipped my mind post rollout as other things came up.

smartapp-developers-guide/scheduling.rst
pfrybar: I mentioned it in the note section: "Similar to the state, only data that can be serialized can be passed to the handler." I can call it out a little more clearly though (i.e. Strings/Numbers/Booleans/Arrays/Maps).

jimmyjames:
This looks good, thanks @pfrybar! We should add this information in some other places as well like the reference docs, and I'm also working on a change to greatly expand on how the json serialization and deserialization works wrt object types. I'll take care of merging this when the related PR goes through, as well as adding any additional information. Thanks!

smartapp-developers-guide/scheduling.rst
charliek: Indeed you did.  Excellent.

charliek:
This looks accurate to me. I'll defer to @jimmyjames on the merging.

jimmyjames:
Closing in favor of https://github.com/SmartThingsCommunity/Documentation/pull/405 which reuses some new docs dealing with JSON serialization, as well as adds it to the quick reference docs.

PR #384, opened 2016-08-30
DOCS-327 Adding ability to consume the capability files

unixbeast:
Adding ability to consume the capability XML and property files in the Docs build.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -14,8 +14,55 @@
 
 import sys
 import os
+import xml.etree.ElementTree as ET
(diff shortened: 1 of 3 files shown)


jimmyjames:
Will merging this have any negative impacts on the build on RTD today? Other than that concern looks good as a first step.

unixbeast:
i'm not sure. I did add a dependency on the python xmltodict module. If the build fails, my guess would be there. But it should be ok because you can use pip to install it locally.

PR #385, opened 2016-09-01
DOCS-332 Clarifying the methods used in the color control example.

unixbeast:


jimmyjames:
Maybe use `hasCommand()` for non-capability methods?

jimmyjames:
👍 for getting rid of the non-capability commands. It might be that there are more standard ways of using this capability (see SmartLighting) that we would want to encourage but for now at least this gets rid of non-capability stuff. 🚢 it

jimmyjames:
👍

PR #396, opened 2016-09-06
occupancy capability was removed when capability definitions moved

jimmyjames:
@unixbeast not sure when this was removed but this is the sort of manual updates we won't have to make soon hopefully!


unixbeast:
💯

PR #412, opened 2016-09-27
[DOCS-343] - runEveryXMinutes/hours being added to device executor base

jimmyjames:
also added all scheduling methods to device handler reference docs


jimmyjames:
@unixbeast not a huge change but would like a quick review nonetheless

unixbeast:
👍

PR #419, opened 2016-10-06
[DOCS-353] document custom remove button

jimmyjames:
@vlaminck please review for accuracy.

/cc @unixbeast @rajkaramchedu


unixbeast:
👍
@vlaminck that cat though...
(actually I can't stop watching it)

PR #426, opened 2016-10-06
[DOCS-355] document SMS rate limiting

jimmyjames:
@lbredeso you should review this.

/cc @rajkaramchedu after this is merged you'll want to add this to the things we'll want to pull out into separate Rate Limiting documentation.


lbredeso:
👍

PR #428, opened 2016-10-07
DOCS-356 updating XML files and the jinja template

unixbeast:
committing latest XML and property file changes.

also updating the jinja template to reflect those changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ download_capabilities:
 extract_capabilities:
 	@echo
 	@echo "Extracting $(CAPABILITIES_ZIP) into $(STATICDIR)"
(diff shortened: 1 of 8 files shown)


jimmyjames:
The capabilities xml and properties files in the `_static` directory are just placeholders to allow us to test the jinja rendering, right?

unixbeast:
No. they are used to do the actual docs build. The _static dir is populated by the Makefile local build. Anything new in there should be committed and addressed in the change log with the weekly release (when we are fully moved over to generated capabilities)
I do have to do some testing with the -o option (overwrite) vs. -f (freshen) option to unzip however to make sure we are only updating stuff that is changed or new when unzipping.

PR #444, opened 2016-10-18
[DOCS-344] - refactor reference API docs to use getter style

jimmyjames:
In anticipation of the updated `getChildApps()` method, it increasingly seems confusing that we document our getters using the Groovy property syntax shortcut. Presenting them as properties also causes confusion because it can lead developers to think that they can set the values as well (which they can't). 

@unixbeast you should review this please.

/cc @rajkaramchedu as you look at the Time story if you make updates to the reference docs you'll want to get this change (once merged) and follow the same pattern.

### DIFF
--- a/ref-docs/app-state-ref.rst
+++ b/ref-docs/app-state-ref.rst
@@ -20,20 +20,20 @@ Here is a small code snippet that illustrates a potential use case.
 
 ----
 
-dateValue
----------
+getDateValue()
+--------------
 
 The value of this Event, if the value can be parsed to a Date.
 
 **Signature:**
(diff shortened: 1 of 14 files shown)


ref-docs/app-state-ref.rst
```diff
@@ -20,20 +20,20 @@ Here is a small code snippet that illustrates a potential use case.
 
 ----
 
-dateValue
----------
+getDataValue()
```
rajkaramchedu: Is this meant to be getDateValue()? A few other instances below also has this error.

unixbeast:
👍 looks good to me

PR #445, opened 2016-10-18
[DOCS-344] - add docs for getAllChildApps() methods and their derivatives

jimmyjames:
Document the following methods, including which return complete and incomplete installations, versus those that only return complete installs:

* `getAllChildApps()` (on SmartApp)
* `findAllChildAppsByName()` (on InstalledSmartAppWrapper)
* `findAllChildAppsByNamespaceAndName()` (on InstalledSmartAppWrapper)
* `findChildAppByNamespaceAndName()` (on InstalledSmartAppWrapper)
* `getAllChildApps()` (on InstalledSmartAppWrapper)

Also clarify that `getChildApps()` will now only return complete child SmartApp installs.

@chrisfsmith please review, thanks!

### DIFF
--- a/ref-docs/smartapp-ref.rst
+++ b/ref-docs/smartapp-ref.rst
@@ -326,16 +326,66 @@ Returns true if the SmartApp is able to schedule jobs. Currently SmartApps are l
 
 ----
 
+findAllChildAppsByName()
+------------------------
+
+Finds all child SmartApps matching the specified name.
+This includes both "complete" and "incomplete" child SmartApp installs.
+
+**Signature:**
+    ``List<InstalledSmartApp> findAllChildAppsByName(String namespace, String name)``
+
+**Parameters:**
+    `String`_ ``name`` - the name of the SmartApp to find.
+
+**Returns:**
+    A list of :ref:`installed_smart_app_wrapper`, or an empty list if none are found.
+
+**Example:**
+
+.. code-block:: groovy
+
+    def children = findAllChildAppsByName(My Child App")
+    log.debug "found ${children.size()} child apps"
+
+----
+
+findAllChildAppsByNamespaceAndName()
+------------------------------------
+
+Finds all child SmartApps matching the specified namespace and name.
+This includes both "complete" and "incomplete" child SmartApp installs.
+
+**Signature:**
+    ``List<InstalledSmartApp> findAllChildAppsByNamespaceAndName(String namespace, String name)``
+
+**Parameters:**
+    `String`_ ``namespace`` - the namespace of the SmartApp to find.
+
+    `String`_ ``name`` - the name of the SmartApp to find.
+
+**Returns:**
+    A list of :ref:`installed_smart_app_wrapper`, or an empty list if none are found.
+
+**Example:**
+
+.. code-block:: groovy
+
+    def children = findAllChildAppsByNamespaceAndName("somenamespace", "My Child App")
+    log.debug "found ${children.size()} child apps"
+
+----
+
 .. _smartapp_find_child_app_by_name:
 
 findChildAppByName()
 --------------------
 
-Returns the first :ref:`installed_smart_app_wrapper` found as a child of this SmartApp that has the specified name.
-
+Finds a child SmartApp matching the specified name.
+This includes both "complete" and "incomplete" child SmartApp installs.
 
 **Signature:**
-    ``InstalledSmartApp findChildAppByName(String appName)``
+    ``def findChildAppByName(String appName)``
 
 **Parameters:**
     `String`_ ``appName`` - the name of the SmartApp to find.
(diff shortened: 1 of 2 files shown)


jimmyjames:
Added the `DO_NOT_MERGE` label since we need the DM deploy with these changes to go out before merging this.

ref-docs/smartapp-ref.rst
```diff
@@ -326,16 +326,66 @@ Returns true if the SmartApp is able to schedule jobs. Currently SmartApps are l
...
+    `String`_ ``name`` - the name of the SmartApp to find.
+
+**Returns:**
+    A list of :ref:`installed_smart_app_wrapper`, or an empty list if none are found.
+
+**Example:**
+
+.. code-block:: groovy
+
+    def children = findAllChildAppsByName(My Child App")
```
chrisfsmith: `"My Child App"`

ref-docs/smartapp-ref.rst
```diff
@@ -326,16 +326,66 @@ Returns true if the SmartApp is able to schedule jobs. Currently SmartApps are l
...
 --------------------
 
-Returns the first :ref:`installed_smart_app_wrapper` found as a child of this SmartApp that has the specified name.
-
+Finds a child SmartApp matching the specified name.
+This includes both "complete" and "incomplete" child SmartApp installs.
 
 **Signature:**
-    ``InstalledSmartApp findChildAppByName(String appName)``
+    ``def findChildAppByName(String appName)``
```
chrisfsmith: I think you could leave this as `InstalledSmartApp` since the `findAllChild...` don't have `def`

ref-docs/smartapp-ref.rst
```diff
@@ -353,16 +403,78 @@ Returns the first :ref:`installed_smart_app_wrapper` found as a child of this Sm
...
 ----
 
+findChildAppByNamespaceAndName()
+--------------------------------
+
+Finds a child SmartApp matching the specified namespace and name.
+This includes both "complete" and "incomplete" child SmartApp installs.
+
+**Signature:**
+    ``def findChildAppsByNamespaceAndName(String namespace, String name)``
```
chrisfsmith: `InstalledSmartApp` instead of `def`

ref-docs/smartapp-ref.rst
```diff
@@ -353,16 +403,78 @@ Returns the first :ref:`installed_smart_app_wrapper` found as a child of this Sm
...
+**Returns:**
+    A :ref:`installed_smart_app_wrapper`, or null if no child app is found.
+    If multiple child apps are found that match the namespace and name, the first one will be returned.
+
+**Example:**
+
+.. code-block:: groovy
+
+    def child = findChildAppsByNamespaceAndName("somenamespace", "My Child App")
+    log.debug "child app id: ${child.id}"
```
chrisfsmith: Add `?` to `${child?.id}` because the return value could be `null`.

Make the same edit to `findChildAppByNamespaceAndName()`'s example.

ref-docs/smartapp-ref.rst
jimmyjames: It's probably not a big deal, but the signature of `findChildAppByName()` is `def`. I think it could be changed in SmartAppExecutorBase to be `InstalledSmartApp` since that's what actually gets returned, but I was erring on the side of what the actual signature is.

ref-docs/smartapp-ref.rst
jimmyjames: Same comment as above; I'm ok using InstalledSmartApp in the docs I guess since that's what it will return, just wish the actual method signature were that as well...

ref-docs/smartapp-ref.rst
jimmyjames: good catch. Also it looks like that example is wrong since it's calling findChildApps.. not findChildApp. I'll fix that.

chrisfsmith:
`findChildAppByNamespaceAndName` also needs the `?` added to the example.

tslagle13:
@jimmyjames in the data returned from getAllChildApps() is there a way to tell which ones are "complete" and which ones are not?

ref-docs/smartapp-ref.rst
```diff
@@ -326,16 +326,66 @@ Returns true if the SmartApp is able to schedule jobs. Currently SmartApps are l
...
+
 .. _smartapp_find_child_app_by_name:
 
 findChildAppByName()
 --------------------
 
-Returns the first :ref:`installed_smart_app_wrapper` found as a child of this SmartApp that has the specified name.
-
+Finds a child SmartApp matching the specified name.
+This includes both "complete" and "incomplete" child SmartApp installs.
```
tslagle13: do findChildAppByName() and findAllChildAppsByName() both return "complete" and "incomplete"?

chrisfsmith:
@tslagle13 http://docs.smartthings.com/en/latest/ref-docs/installed-smart-app-wrapper-ref.html#installationstate

tslagle13:
Cool, thoughts so, just didn't know if there was some special info inside the wrapper.

ref-docs/smartapp-ref.rst
jimmyjames: It states right in the method description that they return both complete and incomplete installations. Unless I'm just totally missing something...

jimmyjames:
@tslagle13 I'll make another PR showing how to tell if a child app is complete/incomplete.

PR #448, opened 2016-10-24
[DOCS-9] Time Methods Draft Doc--[DO NOT MERGE]

rajkaramchedu:
This is the start of the new draft document for time-related methods. I am just getting started, and this is a work in progress (WIP), and the flow and structure may change before the final version.

### DIFF
--- a/smartapp-developers-guide/time-methods.rst
+++ b/smartapp-developers-guide/time-methods.rst
@@ -1,0 +1,83 @@
+.. _smartapp-time-methods:
+
+Time Methods
+============
+
+A powerful feature of SmartThings platform is the ease with which a user can set up automations by which the home monitors itself and triggers events based on what is detected.
+Such monitoring often entails asking the question: "Is it the right time?" and then based on the answer, perform "Do this, or not," actions.
+
+For example, a SmartApp can turn on a room light when a door is opened but only during certain hours, or wake up the house in the morning at different times based on what day of the week it is.
+
+The SmartThings platform provides a few time methods that can readily be used in a SmartApp to accomplish such automations.
+These time methods support a variety of time-related queries such as, to name a few, get the current time, find out today's date, know the time zone, or find out if a given moment of time is between a preset time-window.
+
+timeOfDayIsBetween()
+^^^^^^^^^^^^^^^^^^^^
+
+One such time-related method is ``timeOfDayIsBetween()``.
+A meaningful home automation with SmartThings is to turn on a room light when the door is opened between certain hours, and do not turn on the light during other times.
+Let's see how this ``timeOfDayIsBetween()`` method comes in handy to set up a SmartApp that accomplishes such an automation.
+
+Refer to the SmartApp code below.
+First we set up the ``preferences()`` section with ``openCloseSensor``, a open/close sensor that detects when the door is opened, and a ``roomLight`` that controls the switch to the room light.
+With the ``fromTime`` and ``toTime`` inputs the user will set up the preferred time-window during which the light should be turn on whenever the door is opened.
+
+.. code-block:: groovy
+
+    preferences {
+
+            section("Select SmartThings") {
+    	    input "openCloseSensor", "capability.contactSensor", title: "Which door?", required: true, multiple: false
+            input "roomLight", "capability.switch", title: "Which room light?", required: true, multiple: false
+      }
+      		section("Turn on between what times?") {
+            input "fromTime", "time", title: "From", required: true
+            input "toTime", "time", title: "To", required: true
+        }
+
+    }
+
+
+Next, we begin watching the door by creating the ``contactHandler`` event handler and have it subscribe to the ``contact.open`` attribute of the ``openCloseSensor`` contact sensor.
+This enables the ``contactHandler`` event handler to be sensitive only to the ``open`` event of the contact sensor, i.e, when the door is opened.
+
+.. code-block:: groovy
+
+    def initialize() {
+        subscribe(openCloseSensor, "contact.open", contactHandler)
+    }
+
+
+In the ``contactHandler()`` implementation below, we ensure our SmartApp performs the following checks:
+
+- Is the door open? No? Then do nothing. (in this specific example we do not care if the door is closed.)
+- If the door is open, then are we within the time-window? No? Then do nothing.
+- If the door is open, *and* we are within the time-window, then turn on the room light.
+
+.. code-block:: groovy
+
+    def contactHandler(evt) {
+
+            // Door is opened. Now check if the current time is within the visiting hours window
+
+               def between = true
+               between = timeOfDayIsBetween(fromTime, toTime, new Date(), location.timeZone)
+
+              if (between) {
+             		def currTime = now()
+               		roomLight.on()
+
+                    }
+              else {
+                roomLight.off()
+
+              }
+    }
+
+The ``timeOfDayIsBetween()`` method returns a Boolean ``true`` if the current time and date, i.e., ``new Date()``, is within the time-window specified by ``fromTime`` and ``toTime`` arguments, and returns ``false`` if the ``new Date()`` value is outside the bounds of ``fromTime`` and ``toTime``.
+
+
+.. note::
+  Note that the ``timeOfDayIsBetween()`` resolves to the minute. This means that, for example, if the ``toTime`` argument to it is, say, ``16:14`` (4:14pm), then the time-window is closed before 4:14:59, and *not* before 4:13:59.
+
+----


smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
+.. _smartapp-time-methods:
+
+Time Methods
```
jimmyjames: Maybe make the title of this document more task-oriented, something like "Working with Time"?

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+============
+
+A powerful feature of SmartThings platform is the ease with which a user can set up automations by which the home monitors itself and triggers events based on what is detected.
+Such monitoring often entails asking the question: "Is it the right time?" and then based on the answer, perform "Do this, or not," actions.
+
+For example, a SmartApp can turn on a room light when a door is opened but only during certain hours, or wake up the house in the morning at different times based on what day of the week it is.
+
+The SmartThings platform provides a few time methods that can readily be used in a SmartApp to accomplish such automations.
+These time methods support a variety of time-related queries such as, to name a few, get the current time, find out today's date, know the time zone, or find out if a given moment of time is between a preset time-window.
```
jimmyjames: Minor, but maybe tighten this sentence up a bit. Perhaps we can remove certain phrases like "to name a few" as a start.

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+
+timeOfDayIsBetween()
```
jimmyjames: Let's make this section title more task-oriented instead of the method name. Something like "Taking Action Within a Time Window" (or something to that effect). That allows the reader to understand the use case, and then understand the APIs involved in solving that particular task, rather than starting at the API level and working backwards.

(this isn't something we do a great job of consistently in the docs today, but is something we've learned and are trying to get better at :))

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+^^^^^^^^^^^^^^^^^^^^
```
jimmyjames: Top-level sections should have `----` instead of `^^^^^` (which would be for subsections).

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+
+One such time-related method is ``timeOfDayIsBetween()``.
```
jimmyjames: I think we can remove this sentence and just dive into the use case.

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+A meaningful home automation with SmartThings is to turn on a room light when the door is opened between certain hours, and do not turn on the light during other times.
+Let's see how this ``timeOfDayIsBetween()`` method comes in handy to set up a SmartApp that accomplishes such an automation.
+
+Refer to the SmartApp code below.
+First we set up the ``preferences()`` section with ``openCloseSensor``, a open/close sensor that detects when the door is opened, and a ``roomLight`` that controls the switch to the room light.
```
jimmyjames: "an open/close sensor"

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+            input "roomLight", "capability.switch", title: "Which room light?", required: true, multiple: false
+      }
+      		section("Turn on between what times?") {
+            input "fromTime", "time", title: "From", required: true
+            input "toTime", "time", title: "To", required: true
+        }
+
+    }
+
+
```
jimmyjames: Need to clean up the formatting/indention here. That's usually caused by pasting code from the IDE, which unfortunately mangles indentation with a mix of tabs and spaces :(

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+- If the door is open, *and* we are within the time-window, then turn on the room light.
+
+.. code-block:: groovy
+
+    def contactHandler(evt) {
+
+            // Door is opened. Now check if the current time is within the visiting hours window
+
+               def between = true
+               between = timeOfDayIsBetween(fromTime, toTime, new Date(), location.timeZone)
```
jimmyjames: Don't we need to handle the case where a location has a null time zone? Unless you've discovered otherwise, we should probably show an example similar to our own SmartApps that will use a time zone from either a user preference or some other default, if the location timeZone is null.

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+
+              if (between) {
+             		def currTime = now()
```
jimmyjames: What is the purpose of the `currTime` variable here?

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+               		roomLight.on()
+
+                    }
+              else {
+                roomLight.off()
+
+              }
+    }
```
jimmyjames: Same formatting/indentation issue with this code block.

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+
+The ``timeOfDayIsBetween()`` method returns a Boolean ``true`` if the current time and date, i.e., ``new Date()``, is within the time-window specified by ``fromTime`` and ``toTime`` arguments, and returns ``false`` if the ``new Date()`` value is outside the bounds of ``fromTime`` and ``toTime``.
```
jimmyjames: If it always returns `true` or `false`, why do we set up a variable (`between`) in the code example and initialize it to true? Couldn't we do `def between = timeOfDayIsBetween(...)`? (Maybe there's a reason or case that we can't, I don't know, just as a reader would wonder that.)

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+
+
+.. note::
+  Note that the ``timeOfDayIsBetween()`` resolves to the minute. This means that, for example, if the ``toTime`` argument to it is, say, ``16:14`` (4:14pm), then the time-window is closed before 4:14:59, and *not* before 4:13:59.
```
jimmyjames: This seems like pretty important information, and a potential gotchya. I had to think about this a bit and read the sentence a couple times to get it though. "resolves to the minute" doesn't immediately make sense to me, because I would think it all is minute-based since user inputs cannot be more granular than a minute. But wording aside, if we can use less words to describe this, and then supplement with a few clear examples, that might be much clearer. I could see even listing a few cases in tabular form, with columns for "start time", "end time", "result". Then we could quickly/clearly see how it works.

I think that information should be in the API reference documentation for `timeOfDayIsBetween()` (either as well as, or just instead of here), since it is pretty important for users to understand that. 

We should also link to the API reference docs from the guide, a good rule-of-thumb is that the first reference to the method is a link to the reference docs, and the remaining usages can use the code literal formatting.

smartapp-developers-guide/time-methods.rst
```diff
@@ -0,0 +1,83 @@
...
+For example, a SmartApp can turn on a room light when a door is opened but only during certain hours, or wake up the house in the morning at different times based on what day of the week it is.
+
+The SmartThings platform provides a few time methods that can readily be used in a SmartApp to accomplish such automations.
+These time methods support a variety of time-related queries such as, to name a few, get the current time, find out today's date, know the time zone, or find out if a given moment of time is between a preset time-window.
+
+timeOfDayIsBetween()
+^^^^^^^^^^^^^^^^^^^^
+
+One such time-related method is ``timeOfDayIsBetween()``.
+A meaningful home automation with SmartThings is to turn on a room light when the door is opened between certain hours, and do not turn on the light during other times.
```
jimmyjames: "meaningful" is a bit subjective; maybe just say "A common automation is to turn on a light when a door is opened, but only between certain hours." Or something a bit more concise.

jimmyjames:
Also this guide will need to be added to the smartapp-developers-guide/index.rst so it is included in the ToC.

rajkaramchedu:
Actually, can we delete this PR? I suspect there are a few "not the latest" issues with this branch on my side, and besides we've addressed all these comments in PRs #457 and #456 and I'd like to not keep this branch.

jimmyjames:
Sure, go ahead and close this PR using the "Close and comment" button here.

rajkaramchedu:
Closing this PR. Refer to PRs #457 and #456.

PR #449, opened 2016-10-25
[DOCS-9] Added nextOccurrence() reference

rajkaramchedu:
Added the reference for nextOccurrence() time-method. At the moment the source is the method definition and implementation in https://smartthings.atlassian.net/secure/attachment/38376/CoreApi.txt and an example code that validates the implementation. However, I haven't worked on discovering any gotchas or ins-and-outs we need to watch for, yet.

### DIFF
--- a/ref-docs/smartapp-ref.rst
+++ b/ref-docs/smartapp-ref.rst
@@ -883,6 +883,49 @@ The :ref:`location_ref` into which this SmartApp has been installed.
 
 ----
 
+.. _smartapp_nextoccurrence:
+
+nextOccurrence()
+----------------
+
+Returns a Date when the time specified in the input occurs next.
+
+**Signature:**
+    ``Date nextOccurrence(String timeString)``
+
+**Parameters:**
+    `String`_ ``timeString`` - An ISO-8601 date string as returned from ``time`` input preferences of the SmartApp.
+
+.. note::
+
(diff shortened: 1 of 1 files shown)


ref-docs/smartapp-ref.rst
```diff
@@ -883,6 +883,45 @@ The :ref:`location_ref` into which this SmartApp has been installed.
...
+nextOccurrence()
+----------------
+
+Returns a Date when the time specified in the input occurs next.
+
+**Signature:**
+    ``Date nextOccurrence(String timeString)``
+
+**Parameters:**
+    `String`_ ``timeString`` - Either an ISO-8601 date string as returned from ``time`` input preferences of the SmartApp, or a simple time string in ``"hh:mm"`` format (“16:25”).
```
jimmyjames: `nextOccurrence(timeString)` calls `timeToday(timeString)`, which will throw an `IllegalArgumentException` if a timeZone isn't specified: https://github.com/PhysicalGraph/data-management/blob/23148119a765726a9b47505278d3dfe5efb851ba/src/groovy/physicalgraph/app/api/CoreApi.groovy#L27

So `nextOccurrence("16:25")` would throw that exception. tl;dr seems like the string must be an ISO-8601 date string as returned from the `time` inputs, and we should also document that it will throw an `IllegalArgumentException` if the time string does not contain time zone.

PR #455, opened 2016-10-27
[DOCS-369] Documenting time window use case

rajkaramchedu:
First draft of documenting executing within a time window. At least two more revisions will be made to this doc: (1) Illustrating the default time zone setting and (2) revising the time window logic table.


jimmyjames:
good to go once the PR title is changed to remove "DO NOT MERGE"

PR #456, opened 2016-10-31
[DOCS-369] Modified the logic table at the end (MERGE AFTER the method is released)

rajkaramchedu:
Modified the logic table at the end to reflect fixes in the timeOfDayIsBetween() method. Note: JIRA DOCS-369 is previously moved to DONE but referencing here since it's the relevant Story.


jimmyjames:
Let's add an entry for the upper boundary too (`16:08:00`) to make that clear, then looks good!

rajkaramchedu:
I replaced all rows with 16:xx with 12:xx. Do you mean 12:32:00? If yes, it's there.

jimmyjames:
oh, yeah totally got everything reversed. My bad. 👍

PR #458, opened 2016-11-01
[DOCS-366] fix up scheduling reference docs

jimmyjames:
This gets the scheduling reference docs consistent across SmartApps and Device Handlers, and adds in the signatures for passing a data map to scheduled handler methods.


unixbeast:
💯 looks good to me

jimmyjames:
I also removed the `overwrite` option from any methods other than `runIn()` or `runOnce()`, since that option is not valid for those other methods.

PR #459, opened 2016-11-01
[DOCS-370] document scheduled executions limit

jimmyjames:
@rappleg please review.

/cc @tslagle13 we will need advance community notice about this.


rappleg:
👍

PR #460, opened 2016-11-02
[DOCS-372] clarify getting latest device state

jimmyjames:
@jasonbio @jackchi can you see if this helps to clarify?

@rajkaramchedu would be good for you to review as well since you have a fresh perspective.

### DIFF
--- a/smartapp-developers-guide/devices.rst
+++ b/smartapp-developers-guide/devices.rst
@@ -70,7 +70,9 @@ Device Attributes
 
 Attributes represent the state of a device. A device that supports the "temperatureMeasurement" capability has a "temperature" attribute, for example.
 
-Attributes have state -  the "temperature" attribute has an associated :ref:`state_ref` object that contains information about the temperature (its value, the date it was recorded, etc.)
+Attributes have state -  the "temperature" attribute has an associated :ref:`state_ref` object that contains information about the temperature (its value, the date it was recorded, etc.).
+
+Attribute data is stored in the SmartThings Cloud and updated when the device reports its status.
 
 ----
 
@@ -78,7 +80,8 @@ Device Commands
 ---------------
 
 Devices may expose one or many commands.
-Commands are the things that devices can do. A switch supports the "on" and "off" commands, that turn the switch "on" and "off", respectively.
+Commands are the things that devices can do.
+A switch supports the "on" and "off" commands, that turn the switch "on" and "off", respectively.
 
 Not all devices have commands.
 Commands typically perform some sort of physical actuation (turn a switch on, or unlock a lock, for example).
@@ -89,14 +92,14 @@ A humidity sensor has nothing to physically actuate, for example.
 Getting Device Current Values
 -----------------------------
 
-You can retrieve information about a device's current state in a few ways.
+Information about the most recently reported device attribute state can be retrieved in two ways:
 
-The ``currentState`` method and ``<attributeName>State`` properties both return a :ref:`state_ref` object representing the current state of this device.
+:ref:`device_current_state` and :ref:`device_attribute_state` return a :ref:`state_ref` object that encapsulates the most recently reported state of the device.
 
 .. code-block:: groovy
 
     preferences {
-        section {
+        section() {
             input "tempSensor", "capability.temperatureMeasurement"
         }
     }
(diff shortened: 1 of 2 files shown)


smartapp-developers-guide/devices.rst
```diff
@@ -201,7 +207,9 @@ All commands can take an optional map parameter, as the last argument, to specif
 
 .. note::
 
-    Because specific devices *can* provide more commands than its supported capabilities, it is possible to have more available commands than the capability declares. As a best practice, you should write your SmartApp to the capabilities specification, and not to any specific device. If, however, you are writing a SmartApp for a very specific case, and are willing to forgo the flexibility, you may make use of this ability.
+    Because specific devices *can* provide more commands than its supported capabilities, it is possible to have more available commands than the capability declares.
+    As a best practice, you should write your SmartApp to the capabilities specification, and not to any specific device.
+    If, however, you are writing a SmartApp for a very specific case, and are willing to forgo the flexibility, you may make use of this ability.
```
jimmyjames: More formatting changes, no content changed.

smartapp-developers-guide/devices.rst
```diff
@@ -113,7 +119,7 @@ The ``currentState`` method and ``<attributeName>State`` properties both return
         log.debug "temperature value as an integer: ${anotherCurrentState.integerValue}"
     }
 
-You can get the current value directly by using the ``currentValue(attributeName)`` and its shortcut, ``current<Uppercase attribute name>``:
+You can get the current value directly by using ``latestValue(attributeName)``, ``currentValue(attributeName)`` and its shortcut, ``current<Uppercase attribute name>`` (these can be used interchangeably; they all do the same thing):
```
jackchi: 👍

jackchi:
💯

smartapp-developers-guide/devices.rst
```diff
@@ -89,6 +90,11 @@ A humidity sensor has nothing to physically actuate, for example.
 Getting Device Current Values
 -----------------------------
 
+.. important::
+
+    The current or latest state for an attribute value is the *most recent value the device has reported to SmartThings.*
```
rajkaramchedu: Minor: I am thinking "...or latest state of an attribute value..." makes more sense than, "for an attribute value."

jimmyjames: yep. It was a test, and you passed. I mean I totally did that on purpose. Will push a fix shortly.

smartapp-developers-guide/devices.rst
```diff
@@ -89,6 +90,11 @@ A humidity sensor has nothing to physically actuate, for example.
...
+    It is not calculated by polling or otherwise directly communicating with the device.
```
rajkaramchedu: Maybe I am making more of it than it is, but this immediately begs the question: so where does the state of the attribute come from? I think the above statement is meant to say, "it is not calculated by polling or otherwise directly communicating with the device, at that instance of the query for the latest state." Is it? My understanding is that the latest state of an attribute value is always sourced from communicating with the device, but then again this raises the question: if you are not asking the device what the status is when I need to know what the status is, then how old is the value? Previous value? Okay, so where was that previous value from? (because the logic also applies to previous query, in that even in the past it would've have communicated directly with the device.).

smartapp-developers-guide/devices.rst
rajkaramchedu: There should an emoji for "me patting myself on the back with a grin on my face" :- )

smartapp-developers-guide/devices.rst
jimmyjames: Device attribute values are stored in the SmartThings cloud. So when you do `device.currentValue("switch")` it will get the most recently reported value for the "switch" attribute. The edge case is if the device has gone offline or the Hub has gone offline, or something else happens that prevents the device from checking in, then you could get a value that is inconsistent with what the actual physical device state is.

rajkaramchedu: My point being that explaining there is cloud involved and that devices check in with their status frequently adds to the clarification.

smartapp-developers-guide/devices.rst
jimmyjames: alt text

jimmyjames:
@rajkaramchedu please review with the changes, thanks

rajkaramchedu:
Great, this reads good! Thanks!

jasonbio:
Looks good, thanks :)

PR #464, opened 2016-11-04
DOCS-374 update template

unixbeast:
Noticed I had to update the template to support required arguments.

Also getting rid of some unnecessary legacy garbage at the end of the template.

@jimmyjames @rajkaramchedu


jimmyjames:
👍

PR #466, opened 2016-11-08
[DOCS-361] document manufacturer and model name on device

jimmyjames:
@workingmonk please review

In `device-ref.rst` there are some organizational changes that don't change content. The only change to look at there is the documentation for `getManufacturerName()` and `getModelName()`.

### DIFF
--- a/device-type-developers-guide/definition-metadata.rst
+++ b/device-type-developers-guide/definition-metadata.rst
@@ -126,12 +126,12 @@ This process is known as a "join" process, or "fingerprinting".
 
 The fingerprinting process is dependent on the type of device you are looking to pair.
 SmartThings attempts to match devices coming in based on the input and output clusters a device uses, as well as a profileId (for ZigBee) or deviceId (for Z-Wave).
-Basically, by determining what capabilities your device has, SmartThings determines what your device is.
+By determining what capabilities the device has, SmartThings determines what your device is.
 
 
 .. _zigbee-fingerprinting-label:
 
-ZigBee Fingerprinting
+ZigBee fingerprinting
 ^^^^^^^^^^^^^^^^^^^^^
 
 For ZigBee devices, the main profileIds you will need to use are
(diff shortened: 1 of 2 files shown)


device-type-developers-guide/definition-metadata.rst
```diff
@@ -183,39 +183,29 @@ The fingerprint will be:
...
-
-The reply will be hexadecimal that you can convert to ascii using the hex-to-ascii converter (we'll be adding a utility method to do this, but in the meantime you can use an online converter like `this one <http://www.rapidtables.com/convert/number/hex-to-ascii.htm>`__):
-
-.. code-block:: groovy
-
-    def refresh() {
-        "st rattr 0x${zigbee.deviceNetworkId} 0x${zigbee.endpointId} 0 4", "delay 200",
-        "st rattr 0x${zigbee.deviceNetworkId} 0x${zigbee.endpointId} 0 5"
-    }
-
```
jimmyjames: I removed this section because the Device Handlers I see in the repos do not do this. Let me know if there is something that needs to be called out for the device's `refresh()` method implementation with regards to the manufacturer name/model.

ref-docs/device-ref.rst
```diff
@@ -346,6 +460,70 @@ The name of the Device set by the user in the mobile application or Web IDE.
...
+Gets the manufacturer name of the device, as specified in the Device Handler's :ref:`fingerprint <device_handler_fingerprinting_best_practices>`.
+If the device was joined using a generic fingerprint, it is whatever the device reported while joining.
+
+Not applicable for cloud or LAN-connected devices (``null`` will be returned).
+
+**Signature:**
+    ``String getManufacturerName()``
+
+**Returns:**
+    `String`_ - the manufacturer name of the device, as specified in the device's fingerprinting string. ``null`` if not specified or applicable for the device.
```
jimmyjames: Yeah, that's in the method description but didn't duplicate it in the return description. Maybe most clear to just make the return description super simple (like "the manufacturer name of the device if applicable, `null` otherwise), and leave the details in the method description itself?

jimmyjames: Updated to make the return value docs more generic and defer details to the method description.

jimmyjames:
@workingmonk I updated the ref docs for `getManufacturerName()` and `getModelName()`, can you take a look?

PR #467, opened 2016-11-08
[DOCS-391] - fix build warnings

jimmyjames:
I also created DOCS-393 to look into some automation around this, so we don't have to do this manually.


unixbeast:
👍

PR #469, opened 2016-11-10
[DOCS-396] styles consistent with developer.smartthings.com

jimmyjames:
Some minor styling improvements:

- nav search area color gradient to be consistent with developers.smartthings.com
- more prominent font size and color for the captions in the nav tree
- consistent font for search text

I also reorganized the toc just a little to tighten things up in the nav tree.

Since overriding styles like this isn't ideal, I've created DOCS-397 to create our own sphinx theme (can just be a fork of the current RTD theme with our modifications to start).

### DIFF
--- a/_static/custom.css
+++ b/_static/custom.css
@@ -14,10 +14,31 @@ body {
     font-family: "Swiss 721 Light" !important;
 }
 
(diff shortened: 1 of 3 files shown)


unixbeast:
👍

PR #471, opened 2016-11-10
DOCS-376 updating template to correct missing content

unixbeast:
updated the template for the case where there is only one command that has attributes.


jimmyjames:
wow. I have to trust you on this one :)

👍 if you promise it will never break and someone else has to fix it ;)

unixbeast:
but of course! it was just copy/paste from the like section above.

PR #474, opened 2016-11-10
[DOCS-398] Added warning in Simulator section

rajkaramchedu:
Moved the note from bottom of simulator section to the top and changed it to warning.


jimmyjames:
👍 and also 😞 at the same time

PR #475, opened 2016-11-11
[DOCS-395] Adding "Working with Time Zones" section 

rajkaramchedu:


jimmyjames:
Can you squash these 2 commits before merging?

rajkaramchedu:
Let me do that.

rajkaramchedu:
Did I squash them right?

jimmyjames:
I still see two commits; there should be one.

rajkaramchedu:
Okay, will fix it. Thanks!

jimmyjames:
in this case it's probably not a big deal since one commit is just a normal merge commit. In general we want only one commit but can make an exception in this case.

PR #477, opened 2016-11-11
Fixing lex warnings

unixbeast:
@jimmyjames @rajkaramchedu


jimmyjames:
nice! Some of these were tricky :smirk:

👍

PR #478, opened 2016-11-12
Add ability to copy code blocks

jimmyjames:
Adds the ability to copy code blocks by clicking a button in the upper-right corner, using clipboard.js.

When the mouse enters a code block, the button appears like this:
image

When hovering over the button:
image

After clicking:
image


unixbeast:
👍

PR #485, opened 2016-11-15
Copy code blocks to clipboard, take 2.

jimmyjames:
This PR addresses issues discovered in #478 when deploying to RTD and rendered in IE:

- Fixes issue with button image not displaying by moving image into background custom CSS property. Verified on test RTD site that this path will work.
- This also fixes issue with IE not rendering SVG image.
- This also fixes issue with image not rendering when served from file system.

@unixbeast or @rajkaramchedu I'd like one of you to pull these changes and verify it works as expected. Thanks!


rajkaramchedu:
I pulled and looked at it, looks great to me.

PR #486, opened 2016-11-15
fix more build warnings

jimmyjames:
This fixes the build warning about `oauth-install-restriction.rst` not being included in any toc tree, by changing it to `txt` file.

@unixbeast @rajkaramchedu


unixbeast:
👍 thanks!

PR #491, opened 2016-11-16
DOCS-381 fixing displayed attribute if there is only one

unixbeast:
@jimmyjames @rajkaramchedu 
this was another case where I just copied how attributes were displayed from above in the file.

### DIFF
--- a/test.rst
+++ b/test.rst
@@ -94,7 +94,14 @@ Capabilities Reference
           {%- endif %}
         {% endfor %}
         {%- else %}
-        {{ attribute['value']['@name'] }}
+        {% if attribute['@type'] == 'ENUM' %}
+        ``{{ attribute['value']['@name'] }}``
+        {%- else %}
+            ``{{ attribute['value']['@name'] }}`` - {{ attribute['value']['@type'] }}
+        {%- endif %}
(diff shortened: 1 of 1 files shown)


test.rst
```diff
@@ -94,7 +94,14 @@ Capabilities Reference
           {%- endif %}
         {% endfor %}
         {%- else %}
-        {{ attribute['value']['@name'] }}
+        {% if attribute['@type'] == 'ENUM' %}
+        ``{{ attribute['value']['@name'] }}``
+        {%- else %}
+            ``{{ attribute['value']['@name'] }}`` - {{ attribute['value']['@type'] }}
+        {%- endif %}
+            {% if properties[referenceName][referenceName+".attr."+attribute['@name']+"."+attribute['value']['@name']+".value"] %}
```
jimmyjames: is this supposed to be indented like this?

unixbeast: hm. supposed to be is a stretch. but yes, the indentation is intentional.
At some point it will probably be worth the effort to try and see if we can get the indentation "correct" and not destroy the layout of the rst.

jimmyjames: ok. As I read it there's an if/else/endif followed by another `if`. Just looked to me like that would be indented the same as the previous if/else/endif.

jimmyjames:
👍

PR #497, opened 2016-11-21
[DOCS-331] Add documentation for ZCL message classes

varzac:
This adds documentation for the ZCL message class that can be used for
generating ZCL Zigbee messages for communicating with devices through
the hub.  This abstracts the different methods of messages between v1
and non-v1 hubs.

@unixbeast  Tom and I discussed the possibility of re-organizing some of this stuff.  I think in general the `zigbee.XXX` methods we want to be static backwards compatible calls.  However, the information about the ZCL classes, while it is available and does allow a few very edge case things that aren't supported by the `zigbee` methods, in general we aren't sure we won't make some changes to the interface in the future.  Would it make sense to re-organize and have those in a different location for more specific documentation that is "use with the understanding that there could potentially be breaking changes."  We've had a lot of headaches with the necessity to continue to support everything we've ever had in the past, and this is mostly behind the scenes stuff.  Alternatively we could not add it to the public documentation (albeit it is here in this PR), and thus not have to worry about it and only bring it up as a possible solution if someone contacts us about needing the additional functionality.

Thoughts?

@tpmanley

### DIFF
--- a/device-type-developers-guide/zigbee-primer.rst
+++ b/device-type-developers-guide/zigbee-primer.rst
@@ -4,18 +4,29 @@ ZigBee Primer
 Before we start, lets take a look at a full ZigBee message as it would look in a SmartThings Device Handler.
 Then we’ll break up the message into its parts and dive into what each part means.
 Make sure you download the ZigBee Cluster Library as a reference for ZigBee message formatting and what is possible for each device.
+You can also see our :ref:`zigbee_ref` for more detailed descriptions of our library methods.
 
-Here is a full command:
-``"st cmd 0x${device.deviceNetworkId} ${endpointId} 8 4 {FFFF 0000}”``
+Here are some full commands:
 
-The 3 Main types of ZigBee Messages
+Set the level of a device
+    ``zigbee.command(0x0008, 0x04, "FE0500")``
 
--  st cmd - SmartThings Command which formats the message as a ZigBee
-   Command
--  st rattr - SmartThings Read Attribute which formats the message as a
-   ZigBee Read Attribute
--  st wattr - SmartThings Write Attribtue which as you guessed formats
-   the message as a ZigBee Write Attribute
+Read the current level (e.g. of a light)
+    ``zigbee.readAttribute(0x0008, 0x0000)``
+
+Write the value 0xBEEF to cluster 0x0008 attribute 0x0010
+    ``zigbee.writeAttribute(0x0008, 0x0010, DataType.UINT16, 0xBEEF)``
+
+Report battery level every 10 minutes to 6 hours if it changes value by 1
+    ``zigbee.configureReporting(0x0001, 0x0021, DataType.UINT8, 600, 21600, 0x01)``
+
+The 4 Main types of ZigBee Messages
+
+-  zigbee.command - A ZigBee command for a given cluster
+-  zigbee.readAttribute - A ZigBee Read Attribute requesting the value of an attribute from a cluster
+-  zigbee.writeAttribute - A ZigBee Write Attribute writing a value to the attribute of a cluster
+-  zigbee.configureReporting - A ZigBee Configure Report that configures a cluster attribute to report changes
+  of a given amount within a certain time period
 
 ----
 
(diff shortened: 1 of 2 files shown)


jimmyjames:
Just to clarify for my own understanding, @varzac you are saying that all the ZCL message documentation added in this PR is subject to breaking changes in the future? Seems we should be transparent about that (and perhaps reorganize as you suggested). Will we provide advance notice and deprecation period for breaking API changes? If they are subject to breaking changes without advance notice, they should be called out as such (seems like these are basically *unsupported* APIs?)

varzac:
@jimmyjames Yes.  Basically what it comes down to is that DTH writers should be able to (with a very few exceptions) be able to use the zigbee library method calls for doing everything they need.  That is the API we want people to use, however, this is what we are using behind the scenes and if they were to run up against one of those edge cases that isn't supported they could use this API to do what they needed.  Calling them unsupported APIs is a good way to think about it, they exist and are accessible, but as they are relatively new on our side, if we run into limitations or decide we need to change them for some reason, we want the flexibility to be able to do so instead of having to worry about continuing to support them while trying to migrate to something new.  Some parts should be documented regardless though as they can be used with the zigbee method calls as well (specifically the `DataType` stuff, which is also very unlikely to change as the constants are dependent on the ZigBee specification).

jimmyjames:
I'm thinking it might be good to add an "important" admonition (`.. important::`) at the start of the new ZLL section to state that those APIs are subject to change without notice, or something to that effect.

But... that feels awkward from a developer's standpoint, doesn't it? If I'm working on something and the docs tell me that there are certain APIs to do it, but that they might break at any time without notice... unless it's in a "beta" stage and moving to being supported, I'd probably just be left thinking WTF?

Is there any kind of plan to move these into a supported state, or the required functionality into the ZigBee library? Just trying to think of a way to communicate how these APIs can be used while being clear that they may break at any time, without being awkward or insulting. 

@unixbeast or @tpmanley feel free to offer opinions as well. I may just not be thinking clearly about this due to my ignorance of the subject matter ;)

varzac:
@jimmyjames This is exactly why I was posting the question here.  I agree that it would be strange as a developer to run into that.  That's why I'm sort of leaning towards leaving it as an undocumented API, and just remove the stuff we are unsure of from the documentation.  If anyone for some reason runs into those limitations (which are strange things that I'm not sure we completely support as a platform anyways), we can work with the device manufacturer to support when necessary.  This gives us the freedom to change as we see fit.

jimmyjames:
@varzac I'm good with that. I think the information is valuable for the reasons stated, so I'd suggest maybe sticking it in confluence for now?

varzac:
OK, I will make some changes to only include what I think should be the public changes and I will move the rest into a confluence doc.

ref-docs/zigbee-ref.rst
```diff
@@ -529,3 +540,170 @@ batteryDefect          isBatteryDefectSet()
...
+
+===================== ======================== ===================
+ZCL Data Type          DataType constant name   ZCL numeric value  
+===================== ======================== ===================
+No Data                NO_DATA                  0x00
+8-bit data             DATA8                    0x08
+16-bit data            DATA16                   0x09
+24-bit data            DATA24                   0x0a
+32-bit data            DATA32                   0x0b
+40-bit data	           DATA40                   0x0c
```
tpmanley: Looks like there is an extra tab on this line

device-type-developers-guide/zigbee-primer.rst
```diff
@@ -4,18 +4,29 @@ ZigBee Primer
...
 Then we’ll break up the message into its parts and dive into what each part means.
 Make sure you download the ZigBee Cluster Library as a reference for ZigBee message formatting and what is possible for each device.
+You can also see our :ref:`zigbee_ref` for more detailed descriptions of our library methods.
 
-Here is a full command:
-``"st cmd 0x${device.deviceNetworkId} ${endpointId} 8 4 {FFFF 0000}”``
+Here are some full commands:
 
-The 3 Main types of ZigBee Messages
+Send a set level command to the device
```
tpmanley: Command id 0x04 is the Move to Level (with On/Off). I suggest stating that explicitly or rewording to just say something like "Set the level of a device".

varzac:
@jimmyjames @unixbeast I will leave it to one of you to merge if you are happy with the changes.  I think both Tom and I are good with what's here now, but let me know if you want any changes made.

jimmyjames:
@varzac and @tpmanley you guys rock. 👍

PR #501, opened 2016-11-24
Fix typo with preferences-and-settings.html link

rfinnie:
Fix typo with preferences-and-settings.html link


jimmyjames:
addressing this with #525

PR #502, opened 2016-11-28
[DOCS-406] Image layout adjustments

rajkaramchedu:


rajkaramchedu:
1) Removed white background of the two images, 
2) Retained the image file names for the new images,
3) Used "scale" parameter in reStructuredText source to reduce rendered image sizes.

Issue #514, opened 2016-12-04
Repository Structure: dirs/files need to be lowercase?


airdrummingfool:
Are the directories and folders in a SmartThings repository required to be lowercase in order to work with the GitHub integration? I didn't find anything mentioning lowercase [in the docs](http://docs.smartthings.com/en/latest/tools-and-ide/github-integration.html#repository-structure), but I had trouble adding devices and smartapps from a repository (using the "Update from Repo" button) until I lowercased my files.

If they are required to be lowercase, I'm happy to submit a PR to fix the docs - but I wanted to check first.

rajkaramchedu:
@airdrummingfool Thanks for alerting on this. It is a convention SmartThings adopted, namely that the directory/folder names must all be lower case and must "match" the namespace and name of the device handler or smart app, i.e., all lower case with non-alphanumeric characters replaced with a dash. For example, i.e. if a smart app has the namespace "My Apps" and the name "My First App" then the path name for it must be `smartapps/my-apps/my-first-app.src/my-first-app.groovy`. We will update the docs to reflect this.

jimmyjames:
Closed with #516

PR #516, opened 2016-12-05
[DOCS-418] Add a note on lowercase directory names

rajkaramchedu:
A 3rd party developer asked the question if the directory names need to be in lowercase for the GitHub integration to work. See here: https://github.com/SmartThingsCommunity/Documentation/issues/514
The answer is yes and this PR is to add the clarification to the docs.

### DIFF
--- a/tools-and-ide/github-integration.rst
+++ b/tools-and-ide/github-integration.rst
@@ -151,6 +151,10 @@ The namespace is typically your GitHub user name.
 When you create a SmartApp or device handler in the IDE, you provide a namespace, which is then populated in the definition method.
 This namespace will be used in the directory structure as shown above.
 
+.. important::
+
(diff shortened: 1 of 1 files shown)


unixbeast:
👍

tools-and-ide/github-integration.rst
```diff
@@ -151,6 +151,10 @@ The namespace is typically your GitHub user name.
 When you create a SmartApp or device handler in the IDE, you provide a namespace, which is then populated in the definition method.
 This namespace will be used in the directory structure as shown above.
 
+.. Important::
```
jimmyjames: does this work? Typically admonitions are in lower-case; probably best to use lower-case for consistency if nothing else.

rajkaramchedu:
@jimmyjames good point. It works, but I changed it to lowercase all the same. Thanks!

jimmyjames:
Looks good, 👍 

I'll let you merge this.

PR #520, opened 2016-12-08
some minor visual changes to capabilities reference template

jimmyjames:
Some minor UX changes to the template:

- Removed the trailing `:` from "Attributes" and "Commands" section headers, since section headers by convention do not have a trailing colon.
- Removed bold styling from "Attributes" and "Commands" since it is already a section header.
- Made "Preferences Reference" its own section for each capability, so it appears in the side navigation along with Attributes and Commands, and looks consistent with them.
- Removed trailing colon from the command name, since the indentation for the definition is enough to notify the reader that the definition follows.

Before:

image

With Changes:

image

I'm open to discussing these changes if people don't like them. @unixbeast also please review in case there is some negative impact with modifying the template as I have that I'm not aware of.

### DIFF
--- a/capabilities-reference.rst
+++ b/capabilities-reference.rst
@@ -88,11 +88,12 @@ Capabilities Reference
     {{ properties[referenceName][referenceName+".description"] }}
     {%- endif %}
     {# Capability preference reference #}
-    **Preferences Reference:**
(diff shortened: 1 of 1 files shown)


unixbeast:
I think you also need to remove the ':' on line 264.

unixbeast:
👍

PR #521, opened 2016-12-09
[DOCS-368] Cron schedule elaboration and other readability edits

rajkaramchedu:
Elaborated cron-specific sections highlighted in community here: https://community.smartthings.com/t/announcement-changes-coming-to-cron-jobs/41656/20

Also edited other sections of the document for better readability.
@tslagle13 like to get your eyes on this, please. Thanks!

### DIFF
--- a/smartapp-developers-guide/scheduling.rst
+++ b/smartapp-developers-guide/scheduling.rst
@@ -5,7 +5,7 @@ Scheduling
 
 SmartApps and Device Handlers often need to schedule certain actions to take place at a given point in time.
 For example, an app may want to turn off the lights five minutes after someone leaves.
-Or, an app may want to turn the lights on every day at a certain time.
+Or, an app may want to turn on the lights every day at a certain time.
 
 ----
 
@@ -14,31 +14,31 @@ Overview
 
 Broadly speaking, there are a few different ways we might want to schedule something to happen:
 
-- Do something after a certain time amount from now.
-- Do something once at certain time in the future.
+- Do something after a certain duration of time from now.
+- Do something once at a certain time in the future.
 - Do something on a recurring schedule.
 
-We'll look at each scenario in detail, and what methods SmartThings makes available to address these requirements.
+We'll look at each scenario in detail, and at the methods SmartThings makes available to address these requirements.
 
 .. note::
-    When using the scheduler APIs, the schedule will be created using the time zone for the installed SmartApp's location.
+    When using the scheduler APIs, the schedule will be created using the time zone of the SmartApp's location.
 
 ----
 
-``runIn()`` - Schedule From Now
+Schedule From Now - ``runIn()`` 
 -------------------------------
 
-A SmartApp may want to take some action in a certain amount of time after some event has occurred.
+A SmartApp may want to take some action within a certain duration of time after some event has occurred.
 Consider a few examples:
 
-- When a door closes, turn a light off after two minutes.
-- When everyone leaves, adjust the thermostat after ten minutes.
+- Turn a light off two minutes after a door closes.
+- Adjust the thermostat ten minutes after everyone leaves.
 - If a door opens and is not shut after five minutes, send a notification.
 
-All these scenarios follow a common pattern: when a certain event happens, take some action after a given amount of time.
+All these scenarios follow a common pattern: when a certain event happens, take some action after a given duration of time.
 This can be accomplished this by using the :ref:`smartapp_run_in` method.
 
-``runIn()`` takes the number of seconds from now to execute, the method to execute, and a map of options (not required).
+The ``runIn()`` method executes a specified handler method after a given number of seconds have elapsed. 
 
 .. code-block:: groovy
     :emphasize-lines: 3
@@ -52,10 +52,10 @@ This can be accomplished this by using the :ref:`smartapp_run_in` method.
         theswitch.off()
     }
 
-By default, if a method is scheduled to run in the future, and then another call to runIn with the same method is made, the last one overwrites the previously scheduled method.
+By default, if a method is scheduled to run in the future, and then if another call to ``runIn()`` with the same method is made, the last one overwrites the previously scheduled method.
 This is usually preferable.
 
-Consider the situation if we have a switch scheduled to turn off after five minutes of a door closing:
+Consider a situation where we have a switch scheduled to turn off after five minutes of a door closing:
 
 - First, the door closes at 2:50 and we schedule the switch to turn off after five minutes (2:55).
 - Then, two minutes later (2:52), the door opens and closes again - another call to ``runIn()`` will be made to schedule the switch to turn off in five minutes from now (2:57).
@@ -87,7 +87,7 @@ So, if you do specify ``[overwrite: false]``, be sure to write your handler so t
 
 ----
 
-``runOnce()`` - Run Once in the Future
+Run Once in the Future - ``runOnce()``
 --------------------------------------
 
 Some SmartApps may need to schedule certain actions to happen *once* at a specific time and date. :ref:`smartapp_run_once` handles this case.
@@ -122,7 +122,7 @@ Like ``runIn()``, you can also specify the overwrite behavior of ``runOnce()``:
 
 ----
 
-``schedule()`` - Run on a Recurring Schedule
+Run on a Recurring Schedule - ``schedule()``
 --------------------------------------------
 
 Often, there is a need to schedule a job to run on a specific schedule.
@@ -133,10 +133,9 @@ SmartThings provides the :ref:`smartapp_schedule` method to allow you to create 
 
 The various ``schedule()`` methods follow a similar form - they take an argument representing the desired schedule, and the method to be called on this schedule.
 
-
 .. note::
 
-    If you call ``schedule()`` with a method that is already scheduled, it will result in the schedule for that method being updated with the new schedule.
+    If a method is already scheduled, and later you call ``schedule()`` with that method, then that method will be executed as per the new schedule.
 
 Schedule Once Per Day
 ^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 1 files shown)


smartapp-developers-guide/scheduling.rst
```diff
@@ -14,31 +14,31 @@ Overview
...
+- Turn a light off two minutes after a door closes.
+- Adjust the thermostat ten minutes after everyone leaves.
 - If a door opens and is not shut after five minutes, send a notification.
 
-All these scenarios follow a common pattern: when a certain event happens, take some action after a given amount of time.
+All these scenarios follow a common pattern: when a certain event happens, take some action after a given duration of time.
 This can be accomplished this by using the :ref:`smartapp_run_in` method.
 
-``runIn()`` takes the number of seconds from now to execute, the method to execute, and a map of options (not required).
+``runIn()`` takes, as arguments, the number of seconds from now to execute, the method to execute, and a map of options (not required).
```
jimmyjames: ", as arguments," seems a little redundant and makes it read a little awkward. I think the sentence can be improved though. Maybe something simple like "Simply specify the number of seconds from now to execute and the name of the method to execute" followed by the code example. Or something like that.

smartapp-developers-guide/scheduling.rst
```diff
@@ -52,12 +52,12 @@ This can be accomplished this by using the :ref:`smartapp_run_in` method.
...
 
-By default, if a method is scheduled to run in the future, and then another call to runIn with the same method is made, the last one overwrites the previously scheduled method.
+By default, if a method is scheduled to run in the future, and then if another call to ``runIn()`` with the same method is made, the last one overwrites the previously scheduled method.
 This is usually preferable.
 
-Consider the situation if we have a switch scheduled to turn off after five minutes of a door closing:
+Consider a situation where we have a switch scheduled to turn off after five minutes of a door closing:
 
-- First, the door closes at 2:50 and we schedule the switch to turn off after five minutes (2:55).
+- First, the door closes now at 2:50 and we schedule the switch to turn off after five minutes (2:55).
```
jimmyjames: I'm confused by the addition of the word "now" here, because it makes me think it is right now?

smartapp-developers-guide/scheduling.rst
```diff
@@ -136,7 +136,7 @@ The various ``schedule()`` methods follow a similar form - they take an argument
 
 .. note::
 
-    If you call ``schedule()`` with a method that is already scheduled, it will result in the schedule for that method being updated with the new schedule.
+    If you call ``schedule()`` with a method that is already scheduled, the method will be updated with the new schedule.
```
jimmyjames: "the method will be updated" confuses me. The method to execute is just a method, it won't be updated. I think what we're trying to say here is that the method will be executed now according to the new schedule.

smartapp-developers-guide/scheduling.rst
```diff
@@ -202,58 +200,86 @@ A cron expression is based on the cron UNIX tool, and is a way to specify a recu
...
-``schedule("48 25 10 ? * MON-FRI", handler)``  Execute ``handler()`` at 10:25 AM Monday through Friday (using a randomly chosen seconds field)
+``schedule("48 25 10 ? * MON-FRI", handler)``  Execute ``handler()`` at 10:25 AM Monday through Friday (using a randomly chosen seconds field of 48)
 ============================================= ===========
 
 .. warning::
 
-    ``*`` means "every", not "any".
-    Don't confuse the two.
+    ``*`` means "every", and ``?`` means "any".
+    Note the difference between the two.
```
jimmyjames: I think the original intent of this warning is that people were using `*` and creating high-frequency schedules that were causing issues, when they really wanted `?` (for any). I think we can clarify that here, so that it's clear why developers should care about the difference.

smartapp-developers-guide/scheduling.rst
```diff
@@ -202,58 +200,86 @@ A cron expression is based on the cron UNIX tool, and is a way to specify a recu
...
 
-Cron syntax is very powerful, but also fairly complex.
-
-We recommend reading through the `Quartz Cron Trigger Tutorial`_ for some basic information and examples using the cron expression format.
-
-We also recommend you test your cron expression before using it in a SmartApp or Device Handler.
-It's easy to create a cron expression that doesn't do what you intended, and often this results in too many scheduled executions, resulting in a wide range of problems.
-The cron expression test tool we use is http://www.cronmaker.com/.
+A cron expression is based on the UNIX cron tool, and is a way to specify a recurring schedule.
+SmartThings supports the CronTrigger class, a variant of the UNIX cron expression as implemented in the Quartz open source job scheduling library.
```
jimmyjames: I think it might cause confusion to say we support the CronTrigger class. I would take that to mean I can use that class, just as I could with any other class. But that class is not available to developers, it just may be what we use under-the-hood.

rajkaramchedu:
@llinder I'd like to get your eyes on this please, thanks!

smartapp-developers-guide/scheduling.rst
jimmyjames: Actually to be more specific, we may not actually use Quartz or CronTrigger anymore. Since we previously did, we tried to maintain the same syntax as CronTrigger, even though under-the-hood the implementation may be different.

smartapp-developers-guide/scheduling.rst
```diff
@@ -184,11 +183,9 @@ Finally, you can pass a Long representing the desired time in milliseconds (usin
 Schedule Using Cron
 ^^^^^^^^^^^^^^^^^^^
 
-Scheduling jobs to execute at a particular time is useful, but what if we want to execute a job at some other interval?
```
charliek: So it might be good to call this out as kind of an advanced topic maybe?  Ideally the `runEveryX` methods would be enough for most people.

smartapp-developers-guide/scheduling.rst
```diff
@@ -202,58 +199,86 @@ A cron expression is based on the cron UNIX tool, and is a way to specify a recu
...
-We also recommend you test your cron expression before using it in a SmartApp or Device Handler.
-It's easy to create a cron expression that doesn't do what you intended, and often this results in too many scheduled executions, resulting in a wide range of problems.
-The cron expression test tool we use is http://www.cronmaker.com/.
+A cron expression is based on the UNIX cron tool, and is a way to specify a recurring schedule.
+SmartThings supports a variant of the UNIX cron expressions as implemented in the Quartz open source job scheduling library.
+
+The cron expression supported by SmartThings is a string of six or seven fields, separated by white space. 
+The *seconds* field is the left most field.
+The below table briefly describes these fields. 
+We strongly recommend reading through the `Quartz Cron Trigger Tutorial`_ for some basic information and examples using the cron expression format.
```
charliek: I don't know that we support the full quartz suite of schedules since data management doesn't use quartz any more.  @rappleg do you know what features the new library we use off hand is?

smartapp-developers-guide/scheduling.rst
```diff
@@ -202,58 +199,86 @@ A cron expression is based on the cron UNIX tool, and is a way to specify a recu
...
+Year         empty, 1970-2099 No
+============ ================ ========
 
 .. note::
 
     High volume cron schedules are encouraged to specify a random seconds field.
-    This helps to avoid a large number of scheduled executions being queued up at the same time.
-
-    It may not be acceptable for every use case, but if it is, use a random second.
+    This helps to avoid a large number of scheduled executions being queued up at the same time. If you can, use a random second.
```
charliek: Should we include an example of this?  Or link to one here if one already exists.

smartapp-developers-guide/scheduling.rst
rappleg: The library we use now is cron-utils[https://github.com/jmrozanec/cron-utils](url) and it has builder support for Quartz syntax as well as Unix and Cron4j.  Quartz is the builder that DM currently uses to make sure all of our scheduling syntax was backwards compatible with the old Quartz plugin.

rajkaramchedu:
@charliek Made revisions per all your feedback, love to get your eyes when you can, thanks!

smartapp-developers-guide/scheduling.rst
```diff
@@ -202,58 +198,95 @@ A cron expression is based on the cron UNIX tool, and is a way to specify a recu
...
+Year         empty, 1970-2099 No       , - * /
+============ ================ ======== =================
+
+**Allowed Wildcards**
+
+    - ``,`` (comma) is used to specify additional values. For example, SAT,SUN,MON in the Day of Week field means “the days Saturday, Sunday, and Monday.”
+    - ``-`` (hyphen) is used to specify ranges. For example, ``5-7`` in the Hours field means “the hours 5, 6 and 7”.
+    - ``*`` (asterisk) is used to specify all values in the field. For example, ``*`` in the Hours field means *every* hour.
+    - ``?`` (question mark) is used to specify any value. For example, ``?`` in the Day of Week field means *regardless* of what the day of the week is.
+    - ``/`` (forward slash) is used to specify increments. For example, ``5/15`` in the Seconds field means “the seconds 5, 20, 35, and 50”.
```
charliek: Be cause we don't allow this in the seconds field you might wand to use minutes as the example here.

smartapp-developers-guide/scheduling.rst
```diff
@@ -202,58 +198,95 @@ A cron expression is based on the cron UNIX tool, and is a way to specify a recu
...
-We also recommend you test your cron expression before using it in a SmartApp or Device Handler.
-It's easy to create a cron expression that doesn't do what you intended, and often this results in too many scheduled executions, resulting in a wide range of problems.
+We recommend that you test your cron expression before using it in a SmartApp or Device Handler.
 The cron expression test tool we use is http://www.cronmaker.com/.
 
 .. note::
 
-    High volume cron schedules are encouraged to specify a random seconds field.
-    This helps to avoid a large number of scheduled executions being queued up at the same time.
+    Cron jobs are only be allowed to run at a minimum of 1 minute intervals.
```
charliek: Maximum of 1 minute intervals?

charliek:
Other than the max vs min comment this looks good to me.

smartapp-developers-guide/scheduling.rst
charliek: I was actually just referring to the example `For example, ``5/15`` in the Seconds field means “the seconds 5, 20, 35, and 50”.`  I'd maybe use minutes as the example here since we will not actually honor running multiple times a minute right now.

However your point about allowed wildcards makes sense.  We do allow `*`, `-`, and `/` in the seconds field, but we don't necessarily honor them.  So not sure how we want to document it.  It is a historical artifact and we don't want people polling that often.

rajkaramchedu: @charliek I get it. Good catch. I think I will use minutes in the example, and remove "/" from Seconds table allowable wildcards, as this is the better way to enforce "no faster than a minute" rule for now. We'll cross that bridge when we get there, if we want to support faster-than-a-minute scheduling. Strictly speaking I should remove "-" from Seconds field also, but we can just leave it there for now.

rajkaramchedu:
@charliek Revised it with just two changes: (1) allowed only "*" character in SECONDS field and (2) clarified that faster-than-minute scheduling is not allowed. This should earn your thumbs up, methinks :- )

charliek:
+1 thanks for getting this written up!

PR #524, opened 2016-12-16
[DOCS-389] Adding Capabilities and Devices section - WIP

rajkaramchedu:
Initial draft as a placeholder. 

@bflorian I am wondering when should we be documenting _additionalPermissions()_ in API reference. I understand we are still in phase 1 and we are just logging the SAs that do not explicitly request permission. However, documenting _additionalPermissions()_ now would perhaps encourage SA developers to get used to this policy? Thanks!
@unixbeast like to get your eyes on it, please.

### DIFF
--- a/smartapp-developers-guide/simple-event-handler-smartapps.rst
+++ b/smartapp-developers-guide/simple-event-handler-smartapps.rst
@@ -134,6 +134,17 @@ Refer to the `Sunset and Sunrise <http://docs.smartthings.com/en/latest/smartapp
 
 ----
 
+Capabilities and Devices
+------------------------
+
+Different devices can have a same capability. 
+For example, if an ``input`` in the ``preferences`` section requests for ``capability.battery``, then a lock, a smoke detector and a water sensor at that location will be matched, assuming all these three devices have battery capability attribute.  
+However, you can only interact with the *battery capability* of any of these three devices. 
+In this instance, even though the lock device is a match, you cannot interact with the ``lock`` attribute (i.e., cannot ``subscribe()`` to the ``lock`` attribute events).
+
+If you need to interact with the ``lock`` attribute, then the permission for it must be explicitly requested, with ``additionalPermissions()`` method call.  
+
+
 The Event Object
 ----------------
 


smartapp-developers-guide/simple-event-handler-smartapps.rst
```diff
@@ -134,6 +134,17 @@ Refer to the `Sunset and Sunrise <http://docs.smartthings.com/en/latest/smartapp
 
 ----
 
+Capabilities and Devices
+------------------------
+
+Different devices can have a same capability.
```
rajkaramchedu: Yes 👍

smartapp-developers-guide/simple-event-handler-smartapps.rst
```diff
@@ -134,6 +134,17 @@ Refer to the `Sunset and Sunrise <http://docs.smartthings.com/en/latest/smartapp
...
 
+Capabilities and Devices
+------------------------
+
+Different devices can have a same capability. 
+For example, if an ``input`` in the ``preferences`` section requests for ``capability.battery``, then a lock, a smoke detector and a water sensor at that location will be matched, assuming all these three devices have battery capability attribute.  
+However, you can only interact with the *battery capability* of any of these three devices. 
+In this instance, even though the lock device is a match, you cannot interact with the ``lock`` attribute (i.e., cannot ``subscribe()`` to the ``lock`` attribute events).
+
+If you need to interact with the ``lock`` attribute, then the permission for it must be explicitly requested, with ``additionalPermissions()`` method call.
```
jimmyjames: I think `additionalPermissions()` is another callback method that the SA would define, and the platform would call it (when in the lifecycle it gets called, or under what circumstances, is something to probably doc too). Then in the `addiotonalPermissions()` definition the app would use one of the APIs in the `permissions` namespace, right?

jimmyjames:
Perhaps this topic deserves its own page in the SA guide? It seems important enough to warrant its own page, and it will grow as other related changes are made/released.

smartapp-developers-guide/simple-event-handler-smartapps.rst
rajkaramchedu: Correct.

rajkaramchedu:
Eventually yes, it's own SA guide page. I don't feel we will have enough content during initial phases, i.e., we may have to wait until mobile client UI changes are made to warrant a separate guide.

bflorian:
I think we should wait until we turn on live logging (but not failures) of permission violations before releasing this info to users. Otherwise they won't have any way of seeing what works and what doesn't. I'd expect us to be ready to do that in early January if this current release (with internal-only logging) goes out this week as scheduled.

jimmyjames:
I agree we should wait until there is actual impact to developers. I know people have a plan and understanding of the rollout plans as alluded to, it would really help everyone if these were captured somewhere in confluence so we have some idea of the what and when (it's ok if the "when" is in terms of stages and target timelines; doesn't have to be exact). We can then be proactive with our docs instead of reactive (not publishing until released of course, and understanding that a certain degree of stabilization is required before we can even work on docs drafts).

PR #526, opened 2016-12-21
[DOCS-427] Documenting How to Publish SmartApps and Device Handlers

rajkaramchedu:
@JessicaAtha @jimmyjames Please review, thanks!

PR #529, opened 2016-12-31
Fix footnote anchor.

rpwoodbu:
The footnote mark was not linkified. Adding a preceding space fixed it.


jimmyjames:
Good catch! Turns out this was just fixed as part of a larger PR ( #526 ).

PR #534, opened 2017-01-04
[DOCS-429] Replace remaining instances of Device Type Handler with De…

jimmyjames:
…vice Handler


jimmyjames:
@rajkaramchedu please review, thanks!

ref-docs/device-handler-ref.rst
```diff
@@ -511,7 +511,7 @@ Used within the `tiles()`_ method to define the order that the tiles should appe
 device
 ------
 
-the Device object, from which its current properties and history can be accessed. As of now this object is of a difference class than the Device object available in SmartApps. As some point these will be merged, but for now the properties and methods of the device object available to the device type handler are discussed in the example below:
+the Device object, from which its current properties and history can be accessed. As of now this object is of a difference class than the Device object available in SmartApps. As some point these will be merged, but for now the properties and methods of the device object available to the Device Handler are discussed in the example below:
```
rajkaramchedu: Unrelated to this PR. The above "As some point these will be merged..." should be fixed to "At some point..."

rajkaramchedu: Other than the above comment, looks good! There is another correction we talked about, of replacing "smartdevice" with "device handler." Will that be a separate PR? You might as well knock that off in this PR itself. Either way...

jimmyjames:
Good catches @rajkaramchedu, I had just done a quick find & replace ;)

Updated the PR to account for your comments, as well as cleaned up a couple other terminology issues in the tools-and-ide guide.

PR #541, opened 2017-01-14
[DOCS-422] Documenting HubAction API Reference

rajkaramchedu:
This is a document for HubAction API reference. The "hubaction-ref.rst" is the main document. 
@lbredeso @jimmyjames please review
/cc @bflorian @jodyalbritton @workingmonk 

Thanks for all your help!

Raj

### DIFF
--- a/ref-docs/hubaction-ref.rst
+++ b/ref-docs/hubaction-ref.rst
@@ -1,0 +1,101 @@
+.. _hubaction_ref:
+
+=========
+HubAction
+=========
+
+The HubAction class is used to encapsulate a detailed request string while communicating with the devices in your network.
+For example, this request string can be used for search and discovery of devices in your network. 
+HubAction can also be used in a SmartApp or in a Device Handler to communicate with the device.
+
+**Signature:**
+	``HubAction myhubAction = new physicalgraph.device.HubAction(String action, Protocol protocol)``
+
+	``HubAction myhubAction = new physicalgraph.device.HubAction(String action, Protocol protocol, String dni, Map options)``
+
+	``HubAction myhubAction = new physicalgraph.device.HubAction(Map params, String dni = null, [Map options])``
+
+**Parameters:**
+	``String action`` - A string comprised of the request details targeted for the device.
+
+	``Protocol protocol`` - Specific protocol to be used. Default value is ``Protocol.LAN``.
+
+	``String dni`` - Device Network ID of the device. Default value is null. For dni, we recommend using MAC address and not use IP and port numbers.
+
+	``Map options`` - Default value is null. Available options are:
+
+	======== ===========
+	Option   Description
+	======== ===========
+	callback Name of the callback method
+	type	 Type of LAN request. Allowed values are: LAN_TYPE_CLIENT, LAN_TYPE_SERVER. Default value is LAN_TYPE_CLIENT.
+	protocol Allowed values are LAN_PROTOCOL_TCP and LAN_PROTOCOL_UDP and default value is LAN_PROTOCOL_TCP. Note the difference in allowed values of this parameter when used in ``Maps params`` and ``Protocol protocol`` signatures.
+	======== ===========
+
+	``Map params`` - Available parameters are:
+
+	============ ===========
+	Parameter 	 Description
+	============ ===========
+	path 		 Allowed values are any string of the form ``"/somepath"``. Default value is ``"/"``.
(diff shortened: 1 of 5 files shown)


ref-docs/hubaction-ref.rst
```diff
@@ -0,0 +1,102 @@
...
+
+	``String dni`` - Device Network ID of the device. Default value is null. For dni, we recommend using MAC address and not use IP and port numbers.
+
+	``Map options`` - Default value is null. Available options are:
+
+	======== ===========
+	Option   Description
+	======== ===========
+	callback Name of the callback method
+	type	 Type of LAN request. Allowed values are: LAN_TYPE_CLIENT, LAN_TYPE_SERVER. Default value is LAN_TYPE_CLIENT.
```
lbredeso: I did some double checking in the code, and this is a bit confusing, so I wanted to try to clarify what is going on behind the scenes:

The `protocol` here in `options` is as described below for `params` (LAN_PROTOCOL_TCP, LAN_PROTOCOL_UDP, with default of LAN_PROTOCOL_TCP).

ref-docs/hubaction-ref.rst
```diff
@@ -0,0 +1,102 @@
...
+
+	``Map params`` - Available parameters are:
+
+	===========	===========
+	Parameter 	Description
+	===========	===========
+	path 		Allowed values are any string of the form ``"/somepath"``. Default value is ``"/"``.
+	method 		Allowed values are ``"POST"``, ``"GET"``, ``"PUT"`` and ``"PATCH"``. Default value is ``"POST"``.
+	protocol 	Allowed values are LAN_PROTOCOL_TCP, LAN_PROTOCOL_UDP. Default value is LAN_PROTOCOL_TCP.
+	proto 		Default value is ``"HTTP/1.1"`` It is overridden with the *protocol* parameter above.
```
lbredeso: Developers are not able to pass a `proto` parameter to the `params` argument, so it should be omitted.

ref-docs/hubaction-ref.rst
```diff
@@ -0,0 +1,102 @@
...
+	contentType	Allowed values are ``'application/json'``, ``'text/xml; charset="utf-8"'``.
```
lbredeso: Developers are not able to pass a 'contentType' parameter to the params argument, so it should be omitted. The Content-Type value is pulled directly from `headers`, which can be any valid HTTP/1.1 headers. If the Content-Type is not included in `headers`, we attempt to automatically set it using the following rules:

1. If `params.body` is a `Map`, set the Content-Type to be `application/json`
2. Otherwise, set the Content-Type to be `text/xml; charset="utf-8"`

ref-docs/hubaction-ref.rst
```diff
@@ -0,0 +1,102 @@
...
+.. code-block:: groovy
+
+    sendHubCommand(new physicalgraph.device.HubAction("""GET /xml/device_description.xml HTTP/1.1\r\nHOST: $ip\r\n\r\n""", physicalgraph.device.Protocol.LAN, myMAC, [callback: calledBackHandler]))
+
+    ...
+
+    // the below calledBackHandler() is triggered when the device responds to the sendHubCommand() with "device_description.xml" resource
+
+    void calledBackHandler(physicalgraph.device.HubResponse hubResponse) {
+    	log.debug "Entered calledBackHandler1()..."
```
lbredeso: I think this log statement should be "Entered calledBackHandler()..."

rajkaramchedu:
@lbredeso thanks for your detailed review. I fixed all you highlighted, namely: Fixed descriptions for "protocol", "Content-Type" in "headers", and deleted "proto" and "contentType" options. More important, we know that ``Map`` params.protocol currently only supports ``Protocol.LAN`` value, and since ``Protocol protocol`` gets its values from ``Map`` params.protocol, we are describing that for "Protocol.LAN" is the only allowed and default value for ``Protocol protocol`` parameter. I think this captures it, doesn't it?

ref-docs/hubaction-ref.rst
lbredeso: `protocol` still needs to be documented here for `options`, with possible values of LAN_PROTOCOL_TCP and LAN_PROTOCOL_UDP, and with a default of LAN_PROTOCOL_TCP

rajkaramchedu:
@lbredeso Done the changes. Please review, thanks!

PR #543, opened 2017-01-19
[DOCS-426] Adding Whats New Section

rajkaramchedu:
@jimmyjames See what you think of this approach? The intent is this: the "What's New" page will only focus on last changes, highlighting their benefits with a short blurb, and the docs change log is right below it for those who want to delve into details and for some history.

### DIFF
--- a/architecture/index.rst
+++ b/architecture/index.rst
@@ -249,6 +249,7 @@ Accordingly, SmartApps that use web services will run in the cloud also.
 
        myGarageDoor.open()
 
+.. _hubs-and-locations:
 
 Hubs and Locations
 ------------------
--- a/index.rst
+++ b/index.rst
@@ -27,6 +27,12 @@ This documentation is actively being developed and expanded, so check back often
 
 Contents
 --------
+
+.. toctree::
+   :caption: What's New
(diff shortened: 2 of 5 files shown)


jimmyjames:
My initial thought is that having two separate documents is somewhat confusing. I like the expanded explanations around the changes though. Any reason the change log can't be rebranded and the tone of each item is changed to be similar to what's here?

rajkaramchedu:
That could work too. Let me see how that works.

rajkaramchedu:
@jimmyjames made the change; and modified only the latest release text. Do you think we should modify the text for earlier releases too?

jimmyjames:
I like the single page. I wonder if having a caption "What's New" and page titled "What's New" is a bit too redundant? What if the caption is "What's New" and the page is titled "Recent Changes"?

Regarding changing earlier releases, I think that we should change the last ~4 releases or so. No need to go all the way back, but the most recent ones I think are worth updating.

rajkaramchedu:
@jimmyjames Just made the changes, please review.

jimmyjames:
I like it! 👍

PR #548, opened 2017-01-25
[DOCS-445] Fxing "Latest Updates" for January 23, 2017 release

rajkaramchedu:
Minor editorial and link fixes.

### DIFF
--- a/latest-updates.rst
+++ b/latest-updates.rst
@@ -7,7 +7,7 @@ Latest Updates
 January 23 2017
 ---------------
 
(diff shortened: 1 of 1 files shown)


jimmyjames:
Oops I typed "is" instead of "in" :blush:

Something feels off about the phrase "While the documentation already exists IN the <THING> ON how to use the HubAction class..." though. I think it's the "exists in followed by using the phrase "on how to...". Also if we change the link target but don't specify custom text it will be "While the documentation already exists in the Building the Device Type on how to use the HubAction class", which doesn't really give context. Also I think saying "the documentation already exists" is a bit misleading because we're trying to highlight the new reference docs, contrasted to what is already there.

rajkaramchedu:
@jimmyjames Modified the blurb. How's this?

PR #549, opened 2017-01-30
[DOCS-446] Add getLastActivityDate and getStatus to Device

vl4ds:
These methods are currently undocumented - adding docs:
getLastActivity returns the date of the last event with a source of 'Device'

getStatus returns the status of the device - i.e. ONLINE, ACTIVE, INACTIVE.. if the device doesn't have a status then INACTIVE will be returned.

### DIFF
--- a/ref-docs/device-ref.rst
+++ b/ref-docs/device-ref.rst
@@ -464,6 +464,19 @@ The name of the Device set by the user in the mobile application or Web IDE.
 
 ----
 
+getLastActivity()
+-----------------
+
+The date of the last event with a source of ``device``. (i.e. not commands)
+
+**Signature:**
+    ``String getLastActivity()``
+
+**Returns:**
+    `Date`_ - Date of the last event with a source of ``device``.
+
+----
+
 .. _device_ref_manufacturer_name:
 
 getManufacturerName()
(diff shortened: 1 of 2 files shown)


vl4ds:
cc @jimmyjames @rajkaramchedu  
fyi I'm having problems getting make html to work when setting up the docs so I haven't been able to verify this looks fine locally. If either of you have a few minutes on Monday would you mind helping me troubleshoot? Thanks!

ref-docs/device-ref.rst
```diff
@@ -528,6 +541,19 @@ Not applicable for cloud or LAN-connected devices (``null`` will be returned).
...
+getStatus()
+-----------
+
+Get the current status of the Device. If no status is found then ``INACTIVE`` is returned.
+
+**Signature:**
+    ``String getStatus()``
+
+**Returns:**
+    `String`_ - the status of the Device.
```
jimmyjames: Can we include the possible return values if we know them (as mentioned in the description of this PR)?

vl4ds:
Made the suggested changes. Thanks for the review.

jimmyjames:
Awesome. This looks good to me but @jackchi should verify

jimmyjames:
@mortent maybe you can verify since @jackchi appears to be ooo. Thanks!

PR #553, opened 2017-02-07
Use button instead of tooltip for copy code

jimmyjames:
Use a copy button instead of an icon with tooltips. 

Allows us to remove a bunch of tooltip CSS and the UX is more streamlined as a result.

Before:

image

After:

image

### DIFF
--- a/_static/css/custom.css
+++ b/_static/css/custom.css
@@ -108,230 +108,30 @@ div.column {
     background: #ff4d4d;
 }
 
-.highlight .copy-btn {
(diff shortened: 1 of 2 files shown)


rajkaramchedu:
I've seen earlier that the new COPY button on top right hand is always present (not when we hover) and is blocking the code reading. It was prominent enough for me to notice it. Is it still the same?

jimmyjames:
Yes I made the opacity 50% so you could see the ability to copy was there without having to actually hover into the code block.

PR #557, opened 2017-02-09
[DOCS-425] - more user-friendly docs homepage

jimmyjames:
Updates the docs homepage with a card-based grid layout to bring attention to specific areas of interest for the docs. Uses simple responsive HTML/CSS and FontAwesome (included by RTD theme) for icons.


jimmyjames:
Updated just to remove an errant addition to the ref docs that is not part of this commit

PR #558, opened 2017-02-10
[DOCS-451] Preliminary guide for composite devices

rajkaramchedu:
A preliminary draft of composite devices guide. Much of the content must be vetted by the developers. @jimmyjames please review. I will add devs for review after your feedback.


rajkaramchedu:
@bflorian @jimmyjames please review. @bflorian the entire document needs your blessing, including the note section at the bottom. Thanks!

composite-devices/index.rst
```diff
@@ -0,0 +1,130 @@
...
+
+Composite Devices
+=================
+
+Devices such as Hue LAN bridge, AEON Z-Wave SmartStrip, or a Zooz ZEN20 Z-Wave Power Strip have multiple outputs, and each output can be controlled independently.
+For example, a Zooz ZEN20 Z-Wave Power Strip can be used with a separate Thing connected to each of its five outlets and each Thing can have its own SmartApp.
+
+SmartThings categorizes such a multiple-output device as a *composite device.*
+A device is said to be a composite device when it treats each of its outlets as its child device.
+Integrating a composite device into SmartThings platform requires that its Device Handler, the Service Manager Connect SmartApp and the SmartApp support the composite device functionality.
```
jimmyjames: This makes me think that a Service Manager SmartApp and SmartApp are required, but they needn't be (e.g. the Zooz is just a parent and child device handlers)

composite-devices/index.rst
```diff
@@ -0,0 +1,130 @@
...
+
+----
+
+.. _device_handler_for_composite_device:
+
+Device Handler for a Composite Device
+-------------------------------------
+
+When you integrate a composite device into SmartThings, the composite device maintains a parent-child relationship between itself and its outlets.
```
jimmyjames: "outlets" might be a bit overloaded here, how about just "between itself and its child devices"?

composite-devices/index.rst
```diff
@@ -0,0 +1,130 @@
...
+
+	def installed() {
+		createChildDevices()
+		response(refresh() + configure())
+	}
+
+	...
+
+	private void createChildDevices() {
+		state.oldLabel = device.label
```
jimmyjames: maybe add a comment here why this is needed? When I read it I wonder if this is something I should do, but I don't see it used anywhere.

composite-devices/index.rst
```diff
@@ -0,0 +1,130 @@
...
+		for (i in 1..5) {
```
jimmyjames: I assume the 1->5 loop here is because in this case there are always 5 individual outlets, each of which will be its own child device. If that's the case, a comment in the code example might help the reader understand that.

composite-devices/index.rst
```diff
@@ -0,0 +1,130 @@
...
+		addChildDevice("ZooZ Power Strip Outlet", "${device.deviceNetworkId}-${i}", null,[completedSetup: true, label: "${device.displayName} (CH${i})", isComponent: true, componentName: "ch$i", componentLabel: "Channel $i"])
+		}
+	}
+
```
jimmyjames: We should add the `addChildDevice()` API to the Device Handler reference docs and link to it somewhere here, since there's a lot going on and it's not clear what everything is (for example, what is the `null` param for? What supported options in the map are there and what will they be used for?).

composite-devices/index.rst
rajkaramchedu: @jimmyjames The intent was to generalize this statement to include LAN devices. I'll think of rewording it.

composite-devices/index.rst
rajkaramchedu: @jimmyjames You read my mind :- ) This was on my list to do, but after this Story.

rajkaramchedu:
@jimmyjames See if these changes improve it?

composite-devices/index.rst
rajkaramchedu: See #564

rajkaramchedu:
@jimmyjames Added an API ref for `addChildDevices()` method to Device Handler API ref docs, and linked the API ref to composite devices docs here. Removed dependency on #564 (which no longer exists).

jimmyjames:
LGTM but will need @bflorian or someone of his choosing to review to ensure technical accuracy.

composite-devices/index.rst
bflorian: The `addChildDevice()` method accepts all of the same options as it does when called from a SmartApp. As in that case, the `null` argument is the hub ID. In most cases it won't be needed when the parent is a Device, but we wanted to keep the argument signature the same as for SmartApps, and there could possibly be cases where it is used. 

There are three additional options that apply only when the parent is a device, i.e. not for SmartApp parents:
* isComponent -- hides the device from the Things view and doesn't let it be separately deleted. This value is `true` for the ZEN20 and false for Hue.
* componentName -- a way to refer to this particular child. It should be a Java Bean name (i.e. no spaces). It's used to refer to the device in the parent's detail view (and also for OCF compatibility). It's only needed when isComponent is true.
* componentLabel -- the english name (or i18n key) to be used by the UX

composite-devices/index.rst
```diff
@@ -0,0 +1,133 @@
+.. _composite-devices:
+
+Composite Devices
+=================
+
+Devices such as Hue LAN bridge, AEON Z-Wave SmartStrip, or a Zooz ZEN20 Z-Wave Power Strip have multiple outputs, and each output can be controlled independently.
```
bflorian: I'd suggest using the term "component" rather than "output" here, i.e. "have multiple components"

composite-devices/index.rst
```diff
@@ -0,0 +1,133 @@
...
+For example, a Zooz ZEN20 Z-Wave Power Strip can be used with a separate Thing connected to each of its five outlets and each Thing can have its own SmartApp.
+
+SmartThings categorizes such a multiple-output device as a *composite device.*
+A device is said to be a composite device when it treats each of its outlets as its child device.
```
bflorian: Same here  I think, "multiple-component" device and "each of its components" in the last two sentences

composite-devices/index.rst
```diff
@@ -0,0 +1,133 @@
...
+
+In the above example, the method calls, ``parent.childOn(device.deviceNetworkId)`` and ``parent.childOff(device.deviceNetworkId)``, are the means of communication between the parent and the child instances of this composite device. 
+
+Deleting a Composite Device
+---------------------------
+
+Deleting a composite parent device will delete all children devices. 
+For example, deleting the Power Strip itself will delete its outlets as devices from the SmartThings platform.
+
+A separate and independent SmartApp can be used for each of the outlet devices.
```
bflorian: I was initially confused as to what this sentence meant. Maybe say "SmartApps can be configured to control individual outlets as well as the entire power strip"?

rajkaramchedu:
@bflorian thanks a bunch for your review. I've worked in all your comments and here's the amended commit. In specific, I've modified the language in the composite devices doc and modified the ``addChildDevices()`` API ref for Device Handler to include the three options. Thanks!

jimmyjames:
I see the related PRs for using this feature are either open or have been reverted. Do we need to hold on publishing these docs for the time being, or are we ready to go now? @bflorian

PR #564, opened 2017-02-15
[DOCS-453] Modified API ref doc to cover DH

rajkaramchedu:
@jimmyjames Saw that the ``addChildDevice()`` method is in SmartApp API ref docs already. So I modified it to include the Device Handler case, instead of making a copy of it in the DH API Ref section.


rajkaramchedu:
#558 needs this PR to be merged.

PR #569, opened 2017-02-24
[DOCS-458] Style guide for titles

rajkaramchedu:
@jimmyjames Please review.

### DIFF
--- a/contributing/style-guide.rst
+++ b/contributing/style-guide.rst
@@ -1,12 +1,100 @@
-================
-Writing the Docs
-================
+.. _style_guide:
 
-This document outlines how this documentation should be formatted.
+===================
+Writing Style Guide
+===================
+
+When you write for SmartThings platform, your audience should find your documentation readable, interesting and informative.
+To accomplish these goals, we encourage you to stick to our recommended writing style. 
 
 ----
 
-Page Structure
+Titles and headings
+-------------------
+
+Wherever possible, write purpose-driven documentation. 
+This means writing document titles and section headings that state the benefit explicitly. 
+Such titles or headings can be written as either calls to actions or tasks that can be done. 
+This approach makes it easier for the reader to learn how to get her job done. 
+
+**Examples:** 
+
+- Preferred: Writing Your First SmartApp (document title)
+- Avoid: SmartApp Fundamentals (document title isn't purpose-driven)
+- Preferred: Create your own RESTful API (section heading)
+- Avoid: Parent-child SmartApps (document title isn't purpose-driven)
+- Preferred: Combine Multiple Automations (document title)
+
+.. note::
+
+    - A **document title** is the main title of a document page. A document has only one document title. Example: "Writing Style Guide" at the beginning of this page. The document title also appears at the top level in the navigation bar, so it must be short, preferably four to five words or less. 
+    - A **section heading** is the title for an individual section within a document page. Example: "Titles and headings" at the top of this section. A document page can have multiple sections, and hence multiple section headings. 
+
+Avoid framing as questions
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Avoid using questions in document titles and section headings. 
+
+**Example:** 
+
+- Avoid: How does the switch turn on?
+- Preferred: How the switch turns on (section heading)
+
+Avoid italics (emphasis)
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Avoid using emphasis (italics) in document titles or section headings.
+See :ref:`page_structure`.
+
+----
+
+Document titles
+^^^^^^^^^^^^^^^
+
+Use title case, as in “Document Titles” and not “Document titles.” 
+
+**Example:**
+
+- The title of this document, "Writing Style Guide."
+
+What not to capitalize in title
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- as (SmartApp as a Wakeup Device)
+- to (How to Subscribe to Events)
+- on (Trigger on Time)
+- at (Alarm at a Specific Time)
+- of, the (No More Rules of the Game)
+- in (Your First SmartApp in Five Minutes)
+- the (Motion at the Garage Door)
+- a (Three Critical Triggers in a Day)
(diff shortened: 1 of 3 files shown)


contributing/style-guide.rst
```diff
@@ -1,11 +1,92 @@
...
+This means writing document titles and section headings that state the benefit explicitly. 
+Such titles can be written as either calls to actions or tasks that can be done. 
+This approach makes it easier for the reader to learn how to get her job done. 
+
+**Example:** 
+
+- Preferred: Writing your first SmartApp
+- Preferred: Create your own RESTful API with Web Services SmartApps
+- Avoid: Parent-child SmartApps
+- Preferred: Consolidate separate automations with parent-child SmartApps
```
jimmyjames: This reminds me that we need to keep doc titles to a reasonable length - "Consolidate separate automations with parent-child SmartApps" is probably too long. Since the title is used in the left-hand navigation, titles that are this long will probably make the left-hand navigation too "wordy" and visually cluttered.

Also I think examples here might be a bit confusing, because I don't know if the examples are referring to document titles or section headings (and I think we should go with those terms to avoid confusion, instead of "Document titles" and "Section titles."  Probably best to pull examples and into their respective categories instead of here for general-case.

contributing/style-guide.rst
```diff
@@ -1,11 +1,92 @@
...
+
+What to capitalize
+^^^^^^^^^^^^^^^^^^
+
+- is (Kim Jong-un’s Half Brother Is Reported Assassinated)
+- was (He Was Interviewed About Contact With Envoy, Officials Say)
+- then (5 Years of Drought in California, Then the Threat of Flood)
+- up (He Stood Up to the Nazis for Her; They’ve Been Together Ever Since)
+- that (Trump Knew for Weeks That Flynn Had Not Told Truth)
+- with (Etan Patz Trial Ends With Guilty Verdict for Deli Worker)
```
jimmyjames: Can we find some less-eye-catching examples? Let's avoid any high-profile stories or subjects that might evoke a certain immediate response (like Trump/Nazis/Kim Jong-un). Let's just stay away from news headlines altogether :)

contributing/style-guide.rst
```diff
@@ -1,11 +1,92 @@
...
+
+    - A **sentence case** is where you only capitalize the first letter of the sentence. 
+    - A **title case** is where you capitalize first letter of every word of the sentence. 
+
+----
+
+Question marks in titles
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+Avoid using question marks in document titles and section headings.
```
jimmyjames: Let's put this in the general section/info about titles and headers so it's clear it applies to both

contributing/style-guide.rst
```diff
@@ -1,11 +1,92 @@
...
+^^^^^^^^^^^^^^^^^^
+
+- is (Kim Jong-un’s Half Brother Is Reported Assassinated)
+- was (He Was Interviewed About Contact With Envoy, Officials Say)
+- then (5 Years of Drought in California, Then the Threat of Flood)
+- up (He Stood Up to the Nazis for Her; They’ve Been Together Ever Since)
+- that (Trump Knew for Weeks That Flynn Had Not Told Truth)
+- with (Etan Patz Trial Ends With Guilty Verdict for Deli Worker)
+
+Section titles within the document
```
jimmyjames: How about "Section headings"?

rajkaramchedu:
@jimmyjames Revised. Please check. Thanks!

contributing/style-guide.rst
```diff
@@ -1,11 +1,98 @@
...
+    - A **section heading** is the title for an individual section within a document page. Example: "Titles and headings" at the top of this section. A document page can have multiple sections, and hence multiple section headings. 
+
+Question marks
+^^^^^^^^^^^^^^
+
+Avoid using question marks in document titles and section headings. 
+
+**Example:** 
+
+- Preferred: How does the switch turn on (section heading)
```
jimmyjames: Shouldn't we just avoid questions in section headings altogether? I guess if I see a sentence that is a question, but doesn't have a question mark, it is confusing.

contributing/style-guide.rst
```diff
@@ -1,11 +1,98 @@
...
+- Avoid: How does the switch turn on?
+
+Italics (emphasis)
+^^^^^^^^^^^^^^^^^^
+
+Avoid using emphasis (italics) in document titles or section headings. Use ``html`` tags ``h1, h2, h3, h4, h5 and h6`` or similar strong emphasis styles instead.
```
jimmyjames: Since reStructuredText implements titles/headings into various `h1,h2` etc. tags, Probably best not to add confusion about using those HTML tags. Instead, just don't use italics or bold in titles or headings - rST will handle that for you.

contributing/style-guide.rst
rajkaramchedu: I think question-type headings are fine, unless we want to be strict about it. For example, from today's Economist website: "Why taxing robots is not a good idea" and plenty others when you see it there. And New York Times today, "What It Was Like Onstage During the Oscars 2017 Best Picture Mistake" and so on.

rajkaramchedu: Economist is strict and doesn't allow "?" puncutuation. Times appears to be a bit inconsistent.

contributing/style-guide.rst
rajkaramchedu: I will remove the h1, h2...suggestion but not mention rST because these style guidelines should be applicable whether markdown or Word (aghast!) or rST :- )

contributing/style-guide.rst
jimmyjames: Right but "Why taxing robots is not a good idea" is not a question. "How does the switch turn on" is a question ("How the switch turns on" is not).

contributing/style-guide.rst
jimmyjames: Ok. There is info later in this document I believe that discusses what rST heading formats to use and when (maybe can link to it if it makes sense).

rajkaramchedu:
@jimmyjames Revised as per our talk. Also revised the entire style guide for titles and section headings. Please check, thanks!

PR #573, opened 2017-02-27
[DOCS-458] Style updates for titles and section headings only

rajkaramchedu:
Updated styles for titles and section headings only for three guides: 
1) Architecture, 
2) Rate Limits and 
3) Publishing Code.

@jimmyjames Please check. Also I went with "cloud" instead of "Cloud," and "Web services" instead of "Web Services," as a trial before we decide which way to standardize.

### DIFF
--- a/architecture/index.rst
+++ b/architecture/index.rst
@@ -18,7 +18,7 @@ This allows you to develop an application for a type of device, regardless of th
 All of the code that developers can write on our platform is written in Groovy, which is a dynamic, object-oriented language built for the Java platform.
 You can learn more about Groovy on the :ref:`groovy-basics` page.
 
-Big Picture
+Big picture
 -----------
 
 .. TODO: I think we need a nicer looking picture. (Jesse O'Neill-Oine)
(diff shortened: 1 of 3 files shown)


architecture/index.rst
```diff
@@ -251,7 +251,7 @@ Accordingly, SmartApps that use web services will run in the cloud also.
 
 .. _hubs-and-locations:
 
-Hubs and Locations
+Hubs and locations
```
jimmyjames: Actually I think Location(s) can be considered a proper noun, since they refer to a specific kind of SmartThings object. Point being, I think we should capitalize this here.

publishing/index.rst
```diff
@@ -8,14 +8,14 @@ Publishing for distribution requires you to submit your SmartApp or Device Handl
...
-For Yourself
+For yourself
 ------------
 
 When you publish for yourself, your SmartApp or Device Handler is available *only to your account* on your SmartThings mobile app.
 
 .. _ensure-proper-location:
 
-Ensure Proper Location
+Ensure proper location
```
jimmyjames: Same comment here about location as above

rajkaramchedu:
@jimmyjames Thanks for catching that "location/Location." Here is the revised amended commit.

PR #575, opened 2017-03-01
[DOCS-458] Style updates for titles and section headings

rajkaramchedu:
Style updates for titles and sections for only the following guides:
1) Toos and IDE
2) SmartApps
3) Web Services SmartApps

@jimmyjames please review, thanks!


smartapp-developers-guide/state.rst
```diff
@@ -1,7 +1,7 @@
 .. _storing-data:
 
-State - Storing Data
-====================
+State--Storing Data
```
jimmyjames: Not a big deal, but I think using a hyphen here is probably best avoided altogether, since the title really doesn't represent a break in thought or a range or any other common hyphen usages. Maybe just "Storing Data With State" instead?

rajkaramchedu:
@jimmyjames Agreed on State doc title. Made changes. Please check.

jimmyjames:
looks good 👍

PR #580, opened 2017-03-02
Release notes for 02 march 2017 docs release

jimmyjames:
@rajkaramchedu please review

### DIFF
--- a/latest-updates.rst
+++ b/latest-updates.rst
@@ -4,6 +4,20 @@
 Latest Updates
 ==============
 
+March 02 2017
+-------------
+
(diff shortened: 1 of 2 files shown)


rajkaramchedu:
Don't remember what we concluded, but are we saying "capabilities" and "attributes" or "Capabilities" and "Attributes" (i.e., treating these terms as proper nouns)? Other than this, aok!

jimmyjames:
Not sure if we concluded that yet - I would lean to "Capability" as a proper term, but not attribute or commands. I'll need to think about it a little more before committing to anything, but good to bring it up

PR #597, opened 2017-03-10
[DOCS-463] Revised docs to update glossary terms

rajkaramchedu:
@jimmyjames I think this first scrub looks ok to me but I am sure I missed a few, please check.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,6 @@
 .DS_Store
 _build
 _static/capabilities.zip
+.firebaserc
+database.rules.json
+firebase.json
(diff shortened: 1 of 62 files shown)


rajkaramchedu:
@jimmyjames I went over the doc again with an eye on Events and I recommend keeping it as it is. The title case Event occurrences here refer to SmartThings Events and I didn't see any clear cut deviation from it. Fine tooth comb will help figure out if there are deviations, but I feel we don't need that at this time. What do you think? Also resolved merge conflicts.

PR #598, opened 2017-03-21
[DOCS-465] - doc new tile color standards and update all examples and imgs

jimmyjames:
@rajkaramchedu would be a good idea if you could pull this into your local env and build it to ensure I didn't miss an img file or anything. Thanks


rajkaramchedu:
Looks good, @jimmyjames

rajkaramchedu:
@jimmyjames I am wondering if we should write "Multi-attribute" or "Multi-Attribute" in section headings and in the text. "Multi-attribute" feels more right. Thoughts?

jimmyjames:
I'm fine with that. Open a jira and we can do it in its own PR

PR #602, opened 2017-04-06
[DOCS-466] Guide for Device Health

rajkaramchedu:
@jackchi @mortent @mckeed @jimmyjames This is a new guide for Device Health. Please review, thanks!


jimmyjames:
Looks like a good start, I like the distinction between tracked and untracked devices. I won't get into formatting/organization feedback yet, but I do have some questions after reading through this. Sorry for the sheer volume of questions; this is a pretty complicated and nuanced topic for developers ;)

* Is the `checkInterval` formula 2.2 * reporting interval? Always or just for certain types of devices?
* What are the requirements for implementing the `ping()` command? I assume it must do _something_ that causes the device to check in, but are there certain strategies for various protocols/types of devices?
* Where to send the enroll event (either `checkInterval` or `DeviceWatch-Enroll`)? Is it accurate that `checkInterval` should be sent in `updated()` but `DeviceWatch-Enroll` in `installed()` as it shows here? Why not have both in a `initialize()` method that's called by both `installed()` and `updated()`?
* Is the note about "a good practice" to use `checkInterval` for for ZigBee and Z-Wave and `DeviceWatch-Enroll` for LAN and cloud really a "good practice", or actually a requirement/expectation of how it should be implemented? Either way a little confusing to see it described only how to do one way, but then hinted that it's possible other ways?
* Not really sure what the diagram is supposed to tell me. Do the terminal positions imply something for how I implement Device Health?
* Are all the options in the `DeviceWatch-enroll` event required? I've seen some that don't include the `hubHardwareId`, for example.
* How can I get the current device health status for a health checked device? Is that what `getStatus()` or `getLastActivity()` are for?
* Minimum `checkInterval` time? I think if a value < 60 seconds is given we just rewrite it to be 60 seconds
* For untracked devices, "Whenever you make a decision on the status of the device" - when should I? What's the recommended approach to keeping the status updated correctly? Every time a command succeeds send an "online" event, and every time the device is unreachable (get back bad response from cloud, or on the LAN), mark it "offline"? Or do we want to retry a certain number of times?

Some general troubleshooting questions/tips might be good to have covered as well:
* How to un-enroll a device? Need to have the end user uninstall the device?
* How to debug a device that has gone offline? Are all events including `DeviceWatch-*` events visible via the device's Events page in the IDE?

I am a little concerned that implementing Device Health, especially for Z-Wave devices, is more in-depth than covered here (of course @mckeed is the best resource for this):
* Seems we might be missing a lot of important info specifically for Z-Wave devices (e.g., devices that don't support configuring a reporting interval, commands sent to "beamable" devices can tie up the Z-Wave network, impacting other Z-Wave devices, so checkInterval is recommended to be 1 hour (per the internal wiki))
* In general are there more specific guidelines for the `checkInterval` for different types of Z-Wave devices?

Somewhat related to the Z-Wave case, but probably more general: Do we need to consider that since `ping()` is a command on the device, it is callable from a SmartApp - any guidelines around calling it? Anything we need to warn about, like calling `ping()` on a recurring frequent schedule? For example, what happens if a Z-Wave device is offline but a SmartApp is calling `ping()` on every HealthChecked device, every minute? Could that cause issues to the Z-Wave network itself (/cc @mckeed)?

_(Also, don't forget to add the Health Check capability documentation as well, and mark it as public :) )_

Sorry for the looooooong response ;)

device-health/index.rst
```diff
@@ -0,0 +1,211 @@
...
+---------------
+For the end-user, Device Health is an opt-in feature now available in SmartThings mobile app.
+In the mobile app find *Device Health* option under the menu (in either Android or iOS version) and toggle it ON.
+This will add all your devices for Device Health check.
+You may have a device that is not yet supported for Device Health check.
+In the *Device Health* screen of SmartThings mobile app, tap on *Learn More* to find the `latest list of supported devices for Device Health checks <https://support.smartthings.com/hc/en-us/articles/214529303-Device-Health>`_.
+
+.. note::
+
+  Device Health is not supported by Hub v1.
```
mortent: Is this true?

rajkaramchedu: We've called it out as such in public already. Until I know otherwise, I am staying consistent with that message.

jimmyjames: Yes, the [support KB article](https://support.smartthings.com/hc/en-us/articles/214529303-Device-Health) explicitly states Device Health is v1 only. If that is not the case, we need to let Jessica know and update it there too (I'm assuming that is the case, however).

PR #604, opened 2017-04-10
[DOCS-468] document child install limits

jimmyjames:
There's some reorganization of the Rate Limits docs in here too. Not really related to the child install limit, but wanted to really make our rate limits clear and easy/fast to find.

@gausnes please review.

/cc @rappleg  @rajkaramchedu


ratelimits/index.rst
```diff
@@ -6,91 +6,121 @@ Rate Limits
...
----------------
+----
 
-No more than 15 SMS messages may be sent to the same number per minute.
+Avoiding rate limits
+--------------------
 
-If this limit is exceeded, no additional SMS messages will be sent until the next minute.
+While SmartThings rate limits are quite high compared to other service platforms, the event-driven nature of SmartThings can result in writing code that may (unintentionally) reach this limit.
+It is important to reason carefully about your code, think of worst-case scenarios, and monitor Live Logging when testing to ensure that you are not being rate limited.
```
rajkaramchedu: Line 118 can be cut down to "Monitor Live Logging when testing to ensure that you are not being rate limited."

ratelimits/index.rst
```diff
@@ -6,91 +6,121 @@ Rate Limits
...
-SMS rate limits
----------------
+----
 
-No more than 15 SMS messages may be sent to the same number per minute.
+Avoiding rate limits
+--------------------
 
-If this limit is exceeded, no additional SMS messages will be sent until the next minute.
+While SmartThings rate limits are quite high compared to other service platforms, the event-driven nature of SmartThings can result in writing code that may (unintentionally) reach this limit.
```
rajkaramchedu: "can result in writing code" phrase doesn't feel right. Perhaps drop "writing" and it reads better?

ref-docs/smartapp-ref.rst
```diff
@@ -224,16 +228,28 @@ Adds a child app to a SmartApp.
...
+
+    ``SizeLimitExceededException`` - If this SmartApp already has the maximum number of children allowed (500).
+
+
 addChildDevice()
 ----------------
 
-Adds a child device to a SmartApp. An example use is in service manager SmartApps.
+Adds a child device to a SmartApp.
+An example use is in service manager SmartApps.
```
rajkaramchedu: Service Manager

ratelimits/index.rst
jimmyjames: I don't think that's sufficient, actually, for a few reasons:
* You may not be able to test with the variety and number of devices that will actually be used.
* Live logging does sometimes drop messages.

We've seen cases where developers are surprised that their code is being rate limited. While we have the ability internally to use logging queries to do more in-depth analysis, it is often related to the design of the code itself that can lend itself to problems.

The point being, developers need to think critically about their code in the context of how many executions/length of execution their code may produce given certain scenarios.

ratelimits/index.rst
jimmyjames: Yeah I agree will change it.

ref-docs/smartapp-ref.rst
jimmyjames: Yes, will change.

ratelimits/index.rst
jimmyjames: That said, the end of that sentence can be reworked to clarify that you are not being or likely to be rate limited.

jimmyjames:
added DO NOT MERGE label; will merge prior to docs release to coincide with platform deploy.

gausnes:
@jimmyjames would we want to merge this now considering max children is currently in eu01 and should be getting to the na shards in the near future.

jimmyjames:
@gausnes yeah didn't want to get it in while only on eu01, hoping the deploy goes forward today and I can merge/release it

PR #605, opened 2017-04-20
[DOCS-469] Composite tiles document

rajkaramchedu:
@bflorian @jimmyjames please review. @bflorian please note the changes in the visual layout of the composite tiles for the Simulated Refrigerator example code. I did this to make it easier for the docs reader to follow the example. I will create a separate PR on the repo where Simulated Refrigerator example resides.


composite-devices/index.rst
```diff
@@ -124,8 +130,113 @@ If you try to delete a composite device from your SmartThings mobile app, then t
...
+
+The simulated refrigerator in this example is a composite device with two components (child devices):
+
+- The simulated main refrigerator (fridge) compartment, and 
+- A simulated freezer compartment.
+
+Each compartment has its own door, its own temperature, and its own temperature setpoint. 
+Each compartment is modeled as a child device of the main refrigerator device.
+
+Copy the following three composite device Device Handler files and create your own three Device Handlers with *From Code* option (see :ref:`create-device-handler`):
```
jimmyjames: If these simulated devices are no deployed to prod, do they need to copy/paste and publish for themselves?

composite-devices/index.rst
```diff
@@ -124,8 +130,113 @@ If you try to delete a composite device from your SmartThings mobile app, then t
...
+
+.. _composite_device_tiles:
+
+Composite Device Tiles
+----------------------
+
+A composite device can be visually represented on SmartThings mobile app by using the method :ref:`childDeviceTile_DH_ref`.
+The example below illustrates how to put together a mobile visual interface on SmartThings mobile app for a simulated refrigerator composite device.
+
+Simulated refrigerator
```
jimmyjames: change the title here to "Example walkthrough" or something like that. If I'm browsing the contents, I want to know where an example is, wouldn't necessarily look for "Simulated refrigerator".

rajkaramchedu:
@jimmyjames please review.

composite-devices/index.rst
```diff
@@ -124,8 +130,113 @@ If you try to delete a composite device from your SmartThings mobile app, then t
...
 
 	- A single SmartApp can control all components together in an all-or-nothing fashion.
+
+----
+
+.. _composite_device_tiles:
+
+Composite Device Tiles
+----------------------
+
```
rajkaramchedu: Done. Review when you can.

composite-devices/index.rst
```diff
@@ -124,8 +130,136 @@ If you try to delete a composite device from your SmartThings mobile app, then t
...
+Child device tiles can be visually pulled together into a composite tile.
+On SmartThings mobile app, such a composite tile represents a rich interface for the display and control of a composite device.
+
+For example, consider a refrigerator composite device that is built with two child components, i.e., the fridge door and the temperature control. 
+Then, by using the method :ref:`childDeviceTile_DH_ref` within the parent refrigerator Device Handler, we can define how the individual child device tiles are arranged visually. 
+See example below:
+
```
jimmyjames: I think it might be confusing to show the parent tiles here without the child tile definition. Since `childDeviceTile()` depends on and references the child tile, we should show both.

rajkaramchedu: The thing is, if we show child tile code also in this intro section, then that becomes the entirety of composite tiles documentation, compressed into intro portion of composite tiles documentation. The way it is now is a teaser enough and leaves something to cover in the rest of the section below it. I was inclined to even remove mentioning the childDeviceTile() from this intro, but left it there; serves the teaser purpose. Nutshell: I think we should leave it as it is. What do you think?

jimmyjames: Here's where I'm coming from - when you say "then that becomes the entirety of composite tiles documentation", I see that as a good thing. Less words, easier-to-find information is good for dev docs. If a reader can get the information they need in 30 seconds and less words instead of 3 minutes and more words, that's great and what we should aim for. If a developer wants to go on and spend 3 minutes reading more in-depth docs, great - but they shouldn't have to if they don't want to.

Typically I and other developers will go to the docs, scan prose quick and look for code example, copy it and change it for my needs, and if that's all I need, I'm done. If I need more info or run into issues I'll come back to the docs as needed and look for info related to my needs. 

So I don't think we should be holding/hiding information trying to get devs to read more words if they don't want to - they'll likely just leave the docs and go find a code example or explanation elsewhere (I know I would :) )

rajkaramchedu:
@jimmyjames please review, and thanks! :- )

PR #608, opened 2017-04-28
[DOCS-470] Add fire-and-forget verbiage to the smart-apps documentation

aaron-miller:
We are going to be releasing a small tweak to the async HTTP client soon that enables SmartApp developers to fire off HTTP requests in a 'fire-and-forget' mode where the response is discarded rather than processed. I've updated the documentation to include details around this.


aaron-miller:
Should we wait on merging this until _after_ the release?

jimmyjames:
I'll merge it right around the release, and then build/deploy the docs with those changes. Thanks for doing this!

PR #615, opened 2017-05-20
[DOCS-467] Generated Groovydocs for Z-Wave Reference

rajkaramchedu:
@jimmyjames @mckeed please review. Note the following:

- If you have only a few minutes to look at this PR, then I recommend you visually compare the Codes (i.e., command class ID) in the existing Z-Wave reference table up top, with this generated groovydocs "Description" column. They should be the same. That's the main change. For more details, read below: 

1) In this PR I excluded from groovydocs all command classes and commands that are not already in the current documentation; in specific these are command classes "switchcolorv1", "switchcolorv2", "switchcolorv3", and "BarrierOperatorV1" and all commands of these classes.

We will need to include these command classes in the groovydocs at a future time.

2) added "package-info.groovy" file, with a one-liner describing command class ID and version, for every command. This goes into the "Description" column of the generated groovydocs. 

3) the following command classes are in the current documentation, but the code files are **not** in the repo (i.e., they didn't show up in the generated groovydocs): "Manufacturer Proprietary", "No Operation", and "Non Interoperable". Moreover, the Code (command class ID) for "Multichannel Association" command class and "MultiInstance Association" command class is the same, i.e., 0x8E.  This, I think, **is an error.**


jimmyjames:
1100+ files is a lot to review...

It would be good to include screenshots of how these docs will look especially in terms of where in the navigation are they, since they are full HTML do they open in a new browser tab, etc. 

Also should discuss the process around updating these - I'm assuming you just generated the docs on your machine for this purpose, but we should have a defined process for getting updates to the docs as library changes are released to production. Is there an accomponying PR to the z-wave repo to add the groovyDoc target? We'd want to have that available and not just on your machine.

Also consider that if the docs contents are moved to the new dev portal at SDC, then these URIs will need to redirect there I think. I'm also curious how the `_static` path resolves in the URL especially when used on RTD and if there's anything we need to worry about there.

rajkaramchedu:
I think what was lost in this thread, my bad, was that back in May I had passed on the Firebase link in Slack for the generated html. Now I'll regenerate that and pass it on.

jimmyjames:
Closing this PR as we work out the best to address this given new requirements; we'll circle back later as needed.

PR #628, opened 2017-07-14
Fixed an issue with line 267's onSwitches.size

adamkempenich:


jimmyjames:
@adamkempenich can you please submit this change against `master` instead of `release`? We only merge latest blessed master to `release`.

Otherwise the changes look good!

PR #640, opened 2017-10-14
Update up-and-running.rst

annieprentice:
This is to make it more clear that live logs will only appear for locations where you are listed as the owner.


rajkaramchedu:
Nice touch @annieprentice :- ) thanks!

PR #644, opened 2018-01-17
remove legacy capabilities docs generation

jimmyjames:
This change effectively removes the capabilities reference from docs.smartthings.com, and points to the new capabilities reference on the developer portal. As the capability schemas continue to evolve, maintaining two separate generation techniques is not feasible. 

@rajkaramchedu please review. We can also consider or try adding a redirect in RTD to the capabilities reference on the new portal (not sure if RTD supports redirects to a different domain, but if desired we can try it).

/cc @Bijnagte


rajkaramchedu:
@jimmyjames Looks good. A couple of thoughts. 

- I hope we are not considering capability to capability redirects.

- I think it might be safer if we don't use any redirects in RTD (what if those redirects fail because of some change on RTD side, etc.). The warning admonition we have now, with the link, works better, I think.

- Instead of removing the current Capabilities reference immediately, some sort of a grace period might be useful, just to ensure we don't catch the community unprepared?

jimmyjames:
1. No, no capability-to-capability redirects.
2. Works for me.
3. There's a warning right now at the top of the page, been there for a few weeks. Not sure if people actually noticed it.

jimmyjames:
@tslagle13 FYI

rajkaramchedu:
@tslagle13 @jimmyjames We could mention a date in the current Capabilities page and remove the content on that date. Perhaps a 30-day notice? Which means after 30 days there will be just the warning admonition with the redirect link, and no content below it.

tslagle13:
I do like the notice. 30 days might be a bit long for something like this. I think 15 days would be sufficient.

jimmyjames:
There already is a notice, but that's fine if a date is added to it. Then @rajkaramchedu you can merge this when you're comfortable.

In the meantime, the capabilities documentation here is not current, and the ability to update these docs does not exist. Just want that to be clear so there are no surprises later.

rajkaramchedu:
I am flip-flopping. On second thoughts I see that since the notice was there since a couple weeks ago, might be just simple to make the change after another week or so? @tslagle13 any objections?

PR #648, opened 2018-04-05
Edited the Account creation URL

Rooks103:
With Account Migration live and the register URL hidden in Prod, figured we'd want to change this.


rajkaramchedu:
@Rooks103 Thanks! I will look into this. These docs are now legacy, so we won't update them as often as we do the new docs. I will get to this soon.

